PR #11, opened 2019-10-09
Monster: Several insects

barsoosayque:
This PR includes:
- Giant Wasp
- Giant Bee
- Giant Locust
- Locust Nymph

And also Krita sources for those sprites.

### Preview

prev


I-am-Erk:
They need to be flipped to face right, or they will register as facing the other way when they're moving. Otherwise I love em

barsoosayque:
Alright, I flipped them and redrew shadows (and also moved files according to a new structure). There might be minor light/shadow issues now, but I don't think they are so bad. Here is a new preview:
prev

I-am-Erk:
Another thought, we may want to change the yellow one to be a wasp, as it looks like a threatening yellowjacket sort of critter, and the last one to be a dermatik. Since bees are relatively nonhostile we might want to make them a more bumblebee look

I-am-Erk:
(side note, we should keep your original sprites in storage somewhere so we can convince someone to code sprites to have left and right facing versions instead of just mirroring)

barsoosayque:
> Another thought, we may want to change the yellow one to be a wasp, as it looks like a threatening yellowjacket sort of critter, and the last one to be a dermatik. Since bees are relatively nonhostile we might want to make them a more bumblebee look

I see. Then please don't merge those boys yet, so I can rework them (tomorrow probably), and add some new ones.

So, the changes:
- Giant bee -> Wasp (not a giant one, since a giant wasp is human sized);
- Recolor a giant wasp to yellow/red/black pallet;
- Draw a giant bee as a some kind of nonhostile bumblebee.

> (side note, we should keep your original sprites in storage somewhere so we can convince someone to code sprites to have left and right facing versions instead of just mirroring)

I noted there is `bg` property in json files. Could it potentialy be used as a layer for a shadow ? Like, mirror foreground image, but keep the background as it is.

barsoosayque:
Oh, I just checked: basic wasp is too small to be this yellow thing on the image. It size is _TINY_. I kinda think that my bee is too small to be a giant wasp.

I-am-Erk:
We can do shadows as a BG layer. I haven't been doing that but it's not a bad idea because it stops the shadow from getting flipped... that said, I don't know if we need to worry about it, since the light source is also going to be flipping with the sprite. You can also have as many BG shadows as you want, so you could keep nice looking specific shadows if you feel they're cooler.

Oh the dermatic *is* tiny, wow.
I would suggest keeping the current wasp sprite as a wasp, then, and maybe just edit the yellowjacket to have rounder eyes, and no stinger: 
image

barsoosayque:
What about something like this ? 

mon_bee_source

I can't remove stinger (I just made it smaller to look more non-hostile), because of the description:
> With a stinger the size of a kitchen knife, this dog-sized insect's black and yellow markings warn you to leave it undisturbed.

I-am-Erk:
Yeah I think that's good now. Are you ready for me to merge this?

barsoosayque:
Yep. Gonna draw another bugs in a separate PR.

PR #18, opened 2019-10-11
Deal with invalid content preventing tileset build

ralreegorganon:
* Fixed some syntax issues with item json
* Moved two items that were not correctly sized into purgatory for now


I-am-Erk:
WOops, my bad on the sprites. Forgot to run them through the gimp first.

PR #26, opened 2019-10-16
Slice carpet, linoleum, and colored wall autotiles

anothersimulacrum:
~~Note, `grey` should be `gray`. I've sliced the tiles correctly, but did not rename the source files.~~ Also fixed
~~Also, I'm not sure that I'm doing the `_no_roof` variants of linoleum correctly~~ I wasn't, am now.

bar
clothing
lmoe
mall
pool


I-am-Erk:
Yeah I think I'll have to dim them

PR #30, opened 2019-10-18
Slice rock floor, metal floor, concrete floors

anothersimulacrum:
ant
forest
house
lab
lmoe
restaurant


I-am-Erk:
Woops, forgot trees don't support transparency

PR #42, opened 2019-10-24
Add sheet_metal_small

snipercup:
Add Add sheet_metal_small. Checking to see if I'm doing it right. Please provide feedback.


I-am-Erk:
Everything seems to be in order.

PR #47, opened 2019-10-25
add-first-aid-wielded

snipercup:
I cannot test it because I don't know how to wear a first aid kit. If adding it like this does not make sense, please tell me how to fix it or just close it.


anothersimulacrum:
This should be wielded, not worn

snipercup:
I committed a wielded version but it's not showing up in my game, what am i doing wrong?

PR #57, opened 2019-10-25
add arming sword

snipercup:
image

The middle one is:
image

The other two are:
image

image

The other two need to have looks_like data in their item definition.

image


snipercup:
Added looks_like to the arming sword variants in the game:
https://github.com/CleverRaven/Cataclysm-DDA/pull/35026

PR #60, opened 2019-10-26
add lighter

snipercup:
I'm only getting the sprite I put in first in the 'fg' tag:
image

What do I need to change to make it random? I did check https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/TILESET.md


I-am-Erk:
Items don't support random sprites.

I think I'd like this one too be a bit smaller before it's added

snipercup:
I guess I should mention the artist name: Whaley the Halloween Apparition
Any changes to the sprite should be communicated to Whaley the Halloween Apparition

snipercup:
Changed to incomplete

PR #61, opened 2019-10-26
add cast iron frying pan

snipercup:
image

also created https://github.com/CleverRaven/Cataclysm-DDA/pull/35043


barsoosayque:
This requires a black bottom outline.

Also, I'd say that a cast iron frying pan may be smaller and still be recognizable.

snipercup:
I guess I should mention the artist name: Whaley the Halloween Apparition
Any changes to the sprite should be communicated to Whaley the Halloween Apparition

snipercup:
Changed to incomplete

PR #71, opened 2019-10-28
Add 223 casing AKA rifle caliber casing

snipercup:
image

Artist: ANickel


snipercup:
Other casings will look like this sprite if the following gets merged:

PR #72, opened 2019-10-28
Add .50 casing AKA large caliber casing

snipercup:
image

Artist: Anickel


snipercup:
I did not make any casing look_like the 762 casing so only the 762 casing will get this sprite. Maybe it should be the sprite for the large .50 round instead?

I-am-Erk:
Yep put it in as whatever is the root large caliber casing

snipercup:
Changed it to .50 BMG casing:
image

PR #73, opened 2019-10-28
Add 9mm casing aka pistol caliber casing

snipercup:
image

Artist: ANickel


snipercup:
Other casings will look like this sprite if the following gets merged:

PR #75, opened 2019-10-29
add air filter

snipercup:
image

Artist: ANickel


snipercup:
Should also add looks_like to the makeshift  airfilter in the cdda repo

I-am-Erk:
Hm, still quite big. I'll add it, but I am going to make a folder for incomplete art and move this and several current entries into it.

PR #78, opened 2019-10-30
Tiles: House furniture

barsoosayque:
I wanted to do more but decided to PR early there is a possibility that I do autotiles wrong (so I ask for validation), and also beta might come soon, and having more autotiles in it is great !

And about sofa autotile: I thought that nobody would like to build giant 3x3 sofas (or even more), so the center sprite is just a pouf (just in case).

This PR includes:
- Bathtub autotiles
- Sofa autotiles


barsoosayque:
> Any sofa shaped with a bend at the right is going to glitch

Is there a way to prevent that ?

> Can you put it into incomplete?

I didn't slice autotiles, so the sprites are located in "scratch".

I-am-Erk:
The only way I've thought of to prevent it is to have the sideways tiles have no backs at all, and all horizontal tiles to face the player. We need a code solution.

PR #87, opened 2019-11-06
Fix: Incomplete items

barsoosayque:
Basically I just shrink most of the items, sometimes changing the projection to isometric. With some items I have to redo shading a bit.

## Preview

preview


I-am-Erk:
For the cylindrical objects, I was specifically going to fix their perspective to be more 3/4 - currently sitting on a counter they look pretty tilted I think. For the rest that's pretty much exactly what I'd planned

barsoosayque:
qw
Good enough ?

PR #93, opened 2020-01-22
Add-nailboard-trap

snipercup:
Artist: Esotericist

image

This is the nailboard and nailboard trap item.


esotericist:
minor nitpick: my name should be lowercase. :)

PR #98, opened 2020-02-05
Add CI builds for PRs and master

ifreund:
This will now run builds on every push to master as well as every opened PR.

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -22,7 +22,7 @@ jobs:
     steps:
       - name: Install Dependencies
         run: |
-          apk add --no-cache musl-dev gcc vips-dev python3-dev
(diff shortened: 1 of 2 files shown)


ifreund:
It now also uploads a zip file containing the results of the build after each run.  This means that you can download a zip containing a built version of the current master after every commit from the actions tab.

.github/workflows/ci_build.yml
```diff
@@ -0,0 +1,56 @@
...
+    branches:
+      - master
+
+name: CI Build
+
+jobs:
+  build:
+    name: CI Build
+    runs-on: ubuntu-latest
+    container: alpine:latest
```
int-ua, line 21: why is the container used?

ifreund: iirc, I used that because the old ubuntu version github provided didn't have either a recent enough libvips or any libvips in its repos.

PR #117, opened 2020-02-27
 added boxer briefs and alt sprites, added briefs, added polo shirt, edited clothes.json accordingly, added cig_butt and alt sprites, marble and alt sprites, spoon, fork, knife_butter and source files

eskoONE:


I-am-Erk:
Since we can't display variant types of items, the variant sprites should probably be put into a separate folder within scratch. Like /items/variants/. Then if variants are ever supported we can get them from there.

barsoosayque:
Maybe as with [lighter sprites](../tree/master/gfx/UltimateCataclysm/pngs_normal_32x32/items/lighter), define multiple variants in json, and when this feature comes, we don't have to change anything ? The first variant is the one that will be drawn in the game iirc.

I-am-Erk:
Almost there! You have a duplicate entry for Polo shirt, one in clothing and one in its own folder. Please remove the one in its own folder and then we should be set

eskoONE:
i thought i had that fixed. sry about that. should be good to go now.

PR #134, opened 2020-05-13
Add animation to BrownLikeBears

Fris0uman:
#### Summary
SUMMARY: Content "Add animated tiles"

#### Purpose of change
- Lava
- Console
- Radio controls
- Landmine
- Portal
- Telepad

#### Describe the solution
Add animated tiles
Move multitile for portal to unused folder, multitile conflicts with animation so until it's fixed I think having the animation is better

#### Testing
preview :
https://imgur.com/mlLM0qs


Fris0uman:
Sounds like a good idea

Fris0uman:
By the way, here's the PR that add the ability to add animations: https://github.com/CleverRaven/Cataclysm-DDA/pull/40342

`weight` is the number of frame the tile will be displayed, it doesn't work if the tile as a multitile definition but other than that it seems to work for everything.

barsoosayque:
There are a bunch of `dekstop.ini` files added in your most recent commit. I just want to make sure that isn't an error, because, I mean there could be literally anything in the scratch directory.

Fris0uman:
It was an error, it should be fixed now.

PR #140, opened 2020-05-24
Retrodays+ Changed all trees to 20x20

Wokko1:
All trees are now 20x20

Please tell me if there's something else that has to be done for this pr


barsoosayque:
Everything seems good. I would like to ping @nphyx, because AFAIR he was working on some RD+ trees too. Is this PR conflicting with what you're working on or is this a collaboration or something else ?

nphyx:
It looks like this is just migrating the trees to 20x20 without changing their sprites? That's all good with me. I was only messing around with the trees, and don't intend to commit them any time soon - they still need a lot of work and that may get taken on by someone else anyway (feesp was working on it too I think).

PR #141, opened 2020-05-26
[Retrodays+] Feesp's Hat update

Fris0uman:
Push hat edits made by Feesp


Wokko1:
I already PR'ed fixed hat sprites to Cleverraven, so those png files are separate fixed sprites of hats.

Fris0uman:
You should have PR it to here too at the same time so we can keep the things in sync.

PR #142, opened 2020-05-29
Organize RetroDays+ tiles

anothersimulacrum:
Also removes some tile copies that didn't do anything.

Tested by running around in game, saw nothing out of the ordinary.
Only moves some UI bits out of `effect/`, `wearables/` out of items (maybe this isn't necessary), and organize the `furniture/` directory into several subdirectories.
`terrain/` and `items/` need the same treatment, but that can come in a separate PR.

See [this branch on my repo](https://github.com/anothersimulacrum/CDDA-Tilesets/tree/split/gfx/RetroDays%2B/pngs_tiles_10x10) for how the directories look now.


barsoosayque:
Is this done by hand ? Please don't tell me this is done by hand.. :sweat_smile:
Well, jokes aside, I was wondering if it possible to auto sort tiles. So I actually would like to know was it a manual labor or is there some sort of script which can sort tiles ?

anothersimulacrum:
If there was a script, I would have PR'D it somewhere and done items and terrains along with this :)

PR #149, opened 2020-06-29
Natural terrain update

barsoosayque:
Not gonna lie, I'm very proud of the results. The focus with this one was definitely "make background tiles *background*". So I tried to make image contrast very low, shallow, but overall contrast tile vs character/monster high. Hopefully, this is only the first part of many to come.

Changes:
- Dirt (plus seasonal);
- Snow (but only shallow);
- Grass (plus seasonal);
- Reskin of tall and VERY tall grass (yes, seasonal);
- Dead grass (of course, seasonal);
- All natural water.

Screenshots:

<details>

image
image
image
image
image
image

</details>


I-am-Erk:
Looking great. I do think we need to make winter dirt white, the borders are a bit strange with brown dirt beneath, but this is an improvement. Did we lose the variability to the dirt? The overlay tile system I had would probably still work with this.

barsoosayque:
> I do think we need to make winter dirt white

Totally forgot about this.

> Did we lose the variability to the dirt?

Yes.. I decided to bring it back in the second iteration since I was worried about how clean it would look like.

PR #155, opened 2020-07-18
Add Ultica cats and kittens with GIMP source

wapcaplet:
This PR adds sprites for all 11 cats to UltiCa (both walking and sitting), as well as kittens for each.

I am not exactly sure what I'm doing, but tried to follow `dogs.json` to define the sprites. I see that some monsters use a "weight" to define multiple sprites, so I used this to define the multiple poses (walking and sitting), with a common shadow sprite that works pretty well for both of them (though ideally they would have distinct shadows).

Preview generated with:
```
tools/generate_preview.py  --input gfx  --output kitties.png \
  --monsters  mon_cat  mon_cat_kitten ....(all the cat names)
```
with a gradient added in gimp

image

Pasted over a screenshot from in game with a few of the dogs for comparison:

image


barsoosayque:
I just noticed this PR, and there are a few things I want to request changes for (primarily shadows, but also colors and edges). What would be the best way to request changes ? @wapcaplet Would you agree to implement them post-merge ?

(Small tip: you can generate previews using a directory, e.g. `--from-path pngs_normal_32x32/mon_cat_types`. It's easier than listing all the names)

wapcaplet:
> I just noticed this PR, and there are a few things I want to request changes for (primarily shadows, but also colors and edges).

I think I see your concern about the edges lacking distinction, especially now that I've pasted a sample onto the game terrain. I'd like to make the edges slightly darker and more pronounced. Also the "golden chonker" looks too large in the sitting pose, and the Bengal is perhaps too spotty.

> What would be the best way to request changes ? @wapcaplet Would you agree to implement them post-merge ?

@barsoosayque I would be glad to do another PR with any changes you suggest. We can discuss them here in the comments, or a separate issue / bug report would be fine too.

barsoosayque:
Alright, no need to make an issue for that.

About sprites:

1. Shadows should be changed to the ellipsis ones, because it will work for both left and right direction. Also it will indicate "monsters" more clearer since all other monsters uses that type of shadow. And even more than that, standardized shadows also standartize sprite position within the tile.
2. (I beg your pardon here, I don't know cat breeds) The gold and black checked cat is hard on eyes and is hard to read as a cat. Lower the contrast to the level of the cat next to it. Golden fluffy tail cat is also have a bit of contrast to its colors. Maybe introduce a middle shade or just water them down a bit.
3. Hard edges, yes. Notice golden rottweiler: although this boi is very bright, he has very contrast dark hard edges. For white hard edges, check out duck sprite. I anticipate that it will be super hard to do, because I couldn't do it very well with old cat sprites. It is generally ok to left very light edges (like you did with siamese cat), but otherwise you should have hard edges with every neutral color.
4. Nitpick, but beeg golden cat is slightly too big. Put em on a diet for a little.

Feel free to respond here or open up a new draft PR !

wapcaplet:
I will be submitting a new PR before long but wanted to share the work in progress.

image

Several of the lighter-colored cats have darkened edges (especially the longhair ones). The golden chonker is not *quite* so chonky now, and her a bit of texture. The spotted Bengal (middle of top row) is looking much better I think.

None of the kittens have been modified yet, but I think adding a simple dark edge on their left and lower sides would help them pop out much better.

All are rendered with the `shadow_32x32_small` shadow ellipse.

PR #156, opened 2020-07-19
Revise cat sprites with stronger outlines

wapcaplet:
Following suggestions from @barsoosayque on PR #155, here are some revisions to the cataclysm kitties.

The differences are kind of subtle, so I've prepared some before-and-after pics, on both grass and pavement.

Before:
image

After:
image

Before:
image

After:
image

*Note* The original "mon_cat" and "mon_cat_alt" sprites (the first two brown/grey cats in each screenshot) were left as-is, except to use the common small-monster shadow.

For reference, the cats are:
- Row 1: Original mon_cats (x2), shorthair (x2), Bengal (x2)
- Row 2: Calico (x2), golden chonker (x2), Devon Rex (x2)
- Row 3: Longhair (x2), Maine Coon (x2), Persian (x2)
- Row 4: Siamese (x2), Sphynx (x2), tabby (x2)
- Row 5: Kittens: shorthair, Bengal, Calico, golden chonker, Devon Rex, longhair
- Row 6: Kittens: Maine Coon, Persian, Siamese, Sphynx, tabby

The extra golden chonker with the red bow was the less-shiny draft referenced by @anothersimulacrum


anothersimulacrum:
I like the original walking golden choker sprite better, personally - it loses the shine in the second.

wapcaplet:
> I like the original walking golden choker sprite better, personally - it loses the shine in the second.

Agreed - I made another draft that restores some of the original sheen, and put a red bow on the previous one for comparison.

anothersimulacrum:
That definitely restores the shine! (There's still less, but I'll have to live with it).

wapcaplet:
After looking at these some more, I think I want to make the edges even more pronounced. I will put this in draft mode while I do another pass.

barsoosayque:
Actually, I don't particularly like the original cats, and honestly would like to see them reworked as well. They are too dark, and do not really resemble cats (they were honestly better before that global style rework). If you interested in it, you're more than welcome to make them more cat-ish, otherwise, please do not see them as a perfect cat form, because they are not !

Calico, Devon Rex and tabby have somewhat weak shadows, other cats are pretty good. What I also would like to see is smoother texture on bengal, and no so distinct colors on Maine Coon (sitting version).

wapcaplet:
Here are the results of the last round of fine-tuning, especially the Bengal, Calico, Devon Rex, Maine Coon, Persian, and tabby.

image

I may revisit the two original cats in a future PR, or we could just remove them - I am not sure what will happen with both those and the black cat being "mon_shorthair" - will the cat change color as it walks around?

PR #159, opened 2020-07-26
Blender renders

int-ua:
Also adds `requirements.txt` to tools library.


int-ua:
Waiting on feedback before continuing

barsoosayque:
`t_guardrail.json` is broken (see CI build details)

int-ua:
Should I include `t_guardrail_bg_dp` and `t_guardrail_hw_air` in JSON?

I-am-Erk:
The only thing of note is that the guardrail should have a seasonal variation where the background changes depending on the season. You can see how I did that in https://github.com/I-am-Erk/CDDA-Tilesets/blob/master/gfx/UltimateCataclysm/pngs_normal_32x32/terrain/t_fence/t_fence.json, it's really just a matter of adding season_whatever to the id and changing t_dirt to t_dirt_season or to snow_shallow

int-ua:
the only other place where guardrails are used is a shipwreck and it doesn't look good but I'm going to postpone fixing it indefinitely for now as it's a rare feature. I've added primitive outlines to cow and chicken and would like this to be merged as is (they are in the unfinished folder anyway) to then rebase my branch on the latest master (hit the same saddle problem)

I-am-Erk:
Really guardrails on a shipwreck represent a code problem, guardrails should be furniture, not terrain. THat's clearly out of scope for now.

PR #160, opened 2020-07-26
Blender renders, second batch

int-ua:
Notice that extinguisher is replacing an older one, please comment.
WIP screenshots
2020-07-27-053701_1920x1080_scrot
2020-07-27-172712_1920x1080_scrot


int-ua:
Looks like readme needs updating, please add these new rules and I'll update my templates.

int-ua:
Current state: 
2020-07-27-172712_1920x1080_scrot

PR #164, opened 2020-07-29
Blender renders: bucket, ceramic mug, machete, solar panel and cell, steel pan

int-ua:


I-am-Erk:
Honestly, I'm fine with all of these except maybe the pan and bucket going into normal_32x32, although ideally the mug at least should go to small_20x20 and have its canvas cropped to suit... I don't object to you skipping that.

int-ua:
I'll crop and move them, np.

PR #170, opened 2020-08-01
zweihander

int-ua:


barsoosayque:
Looks very fantasy to me. And doesn't really resembles a zweihänder in my opinion (including that source 3d model). Proportions are off: grip is too short, blade is too short, the second parrying guard is just enormous. If you can erase the second guard, it might work as a regular longsword.

int-ua:
What does resemble a zweihänder in your opinion? :)
Anyway, I like it as it is, you can either accept it or reject.

I-am-Erk:
It doesn't look too different from historical zweihanders to me

https://en.m.wikipedia.org/wiki/Zweih%C3%A4nder

int-ua:
Oh, I can move it into the incomplete folder, ok?

barsoosayque:
> Oh, I can move it into the incomplete folder, ok?

Yep

int-ua:
done

PR #171, opened 2020-08-04
m107a1, water and wind mills and other fillers

int-ua:


I-am-Erk:
The windmill is a larger image and will need to be uploaded to the 32x64 spritesheet. We don't have an Incomplete sheet for that but it's OK if it's marked just on the spreadsheet.

int-ua:
I think I'm done for today, feel free to merge.

PR #172, opened 2020-08-05
Mounted solar panel filler

int-ua:


I-am-Erk:
Hmm. I don't love this as the furniture version - I think the mounted solar panel should be more like this:
image
If you're able to swing something like that, even just rendering this at an angle for us to later add in the supports, that'd be neat.

int-ua:
Yes, I completely agree, but I don't have time for that right now. This image was just a leftover from previous renders and vehicle solar panels already have a filler.

PR #178, opened 2020-08-09
[BLB] Make smoke field transparent

Fris0uman:
image

I've also included the xcf files since I used layers. I've just put them directly in the same folder as the rest, is that ok?


I-am-Erk:
We use a separate folder for source files in Ultica, but it doesn't hurt anything to keep them all together, you can run blb as you wish

PR #186, opened 2020-08-18
[CR] Templates draft

int-ua:
Folders that can be copied for adding new sprites. Not sure about multitile the most as there is a script already.


templates/json/multitile/multitile.json
```diff
@@ -0,0 +1,60 @@
...
+        "RENAME_ME_end_piece_n",
+        "RENAME_ME_end_piece_w",
+        "RENAME_ME_end_piece_s",
+        "RENAME_ME_end_piece_e"
+      ]
+    },
+    {
+      "bg": [],
+      "id": "unconnected",
+      "fg": [ "RENAME_ME_unconnected", "RENAME_ME_unconnected" ]
```
int-ua, line 57: I've copied it from `f_table`, is it supposed to be like this?

I-am-Erk: If you mean two copies, yes, that tells the game that it should not rotate. Multitiles are assumed to rotate, so the game needs to be told that "all the rotations look the same" basically. It's worth adding a "//" comment to that effect maybe.

PR #188, opened 2020-08-19
Add new white and yellow cursors and dotted line

wapcaplet:
The existing cursor for `x`, `V`, and vehicle heading, as well as the semitransparent dotted line toward items and monsters inspected with `V`, is difficult to see especially at low-resolution zoom levels (25% and 50%).

This PR adds new sprites for white and yellow cursors, and makes the dotted-line sprite solid black and white instead of semitransparent.

- Use all-white cursor for the 'x' (look around) cursor
- Use all-yellow cursor for the 'V' (view monster/item) cursor

New green, yellow, and red cursors sprites are also included, though they are not used for anything yet.

Here is the old `x` cursor at 25%, 50%, and 100% zoom:
image

And the new cursor with the same zoom levels (new dotted line not included):

image

To emphazise the benefits at lower zoom levels, this screenshot compares the old dotted-line and monster-highlight cursor with the new one, using a 25% zoom. The line and cursor are practically invisible with the old sprites, but show up well with the new ones:

image

I considered using a black outline or even a black shadow on the cursor so it would stand out well against white or light-colored backgrounds, but in nearly all cases it looked screwy and unpredictable at the lower zoom levels. A pure white cursor stands out well at all zoom levels, since it's rare for any other sprites to contain pure white. Even against a pure snow background, white is not difficult to see:

image


I-am-Erk:
I'm willing to give 'em a try. If they don't stand out, another option would be to animate them to blink black-white-black-white slowly.

PR #190, opened 2020-08-20
10 new clothes by ACEPLEIADES

int-ua:


barsoosayque (commented):
can you rename folders with space in their names to have underscores instead ?

int-ua:
done

PR #191, opened 2020-08-20
artifacts, cu_pipe

int-ua:
WIP


I-am-Erk:
I'll merge this as is, but it occurs to me that an interesting direction to explore might be rendering artifacts without your pixelizing script, makign them stand out and look strange. (nod to my departed friend and webcomic Erfworld)

PR #193, opened 2020-08-20
more artifacts, tr_portal

int-ua:


int-ua:
I know beads are huge and not good. But moving only them to unfinished seems too tedious so I'll leave them here, ok?

int-ua:
No hard feelings about the portal, can move it to the unfinished or revert.

I-am-Erk:
I am not sure if items support multiple sprites, you may want to test that.

int-ua:
Ok, a bit later. Meanwhile you can select which ones should be culled.

int-ua:
Ok, looks like it's not supported for items, which ones should I delete?

I-am-Erk:
For now I would keep gamma2

PR #213, opened 2020-08-23
34 powders

int-ua:
Source for making more: https://discord.com/channels/552510581161066497/552510581782085654/747173242753515651

After I've made them I realized that they are too big >_<. Maybe I'll try to reduce the size later.

Screenshot_20200823_214547


int-ua:
```sh
$ file *.png | grep gray -B1 -A1
powder_black_gunpowder.png:   PNG image data, 32 x 32, 8-bit colormap, non-interlaced
powder_calcium_carbide.png:   PNG image data, 32 x 32, 8-bit gray+alpha, non-interlaced
powder_chili.png:             PNG image data, 32 x 32, 8-bit colormap, non-interlaced
--
powder_mustard.png:           PNG image data, 32 x 32, 8-bit colormap, non-interlaced
powder_oxidizer.png:          PNG image data, 32 x 32, 8-bit gray+alpha, non-interlaced
powder_peptone.png:           PNG image data, 32 x 32, 8-bit colormap, non-interlaced
```
hmm

int-ua:
`$ convert powder_oxidizer.png png32:powder_oxidizer.png`

I-am-Erk:
I'd like these to be kept in incomplete, if only so that we can go through them and consider the visual size of the piles. Like, a pile of chili powder on the ground should be a lot smaller than a pile of flour.

int-ua:
Done.

PR #252, opened 2020-09-09
empty PNG for hot air

int-ua:
May fix https://github.com/CleverRaven/Cataclysm-DDA/issues/44042


I-am-Erk:
Call the PNG "transparent", we're likely to want it for other fields

int-ua:
Do we still need it when it was fixed upstream? I'm inclined to close this PR unless you want these in.

int-ua:
Should I move it to some more generic directory?

I-am-Erk:
we may need to revisit this some day but for now let's not worry about it

PR #256, opened 2020-09-16
discord submissions 2020-09-16

int-ua:


I-am-Erk:
Can you put Blinky, the geese, and the snakes into the monsters/animals folder? And I think the cats should use the circle shadows, at least until the darker shadows have the correct light direction

PR #261, opened 2020-09-25
discord submissions 2020-09-25

int-ua:


I-am-Erk:
There's a few here I would like to see alterations to before they get added
- chickenbot
- razorclaws
- science bot
- robotfac prototype
These need to either be in an Incomplete tilesheet, or be adjusted to fit the style better before addition.

PR #262, opened 2020-09-26
Fixed compose.py warnings

int-ua:
`cursor.png` was identical to `line_target.png`


int-ua:
Wait, how are there no conflict after #267 was merged?

PR #267, opened 2020-09-30
correct rotation data for vehicle doors in ultica

esotericist:
a bunch of vehicle doors repeated `rotS` in their definitions for the open state
replaced those with `rotE` where appropriate

before:
image

after:
image


int-ua:
https://github.com/I-am-Erk/CDDA-Tilesets/pull/262/commits/c6b22c3f8a8c1f78ee75350d93b551c5c05065c4

PR #278, opened 2020-10-17
Put wingmirrors into centered folder/tilesheet

Fris0uman:
readme in ``pngs_centered_64x64` says:
```
## About this folder
This folder should hold only JSON files, pointing to sprites stored in the large_64x64 folder, to keep from needing an entire separate spritesheet.
```
But the compose script is making an empty tilesheet from it. So I moved the wing mirrors into it so that they are in the right sheet with the right offsets


int-ua:
They were added for making easier to add custom fillers, like https://discourse.cataclysmdda.org/t/text-fallback-fillers-an-alternative-to-ascii-fallbacks/24573

Fris0uman:
This might need some doc shipped with the tileset then, no?

int-ua:
Ok, will you agree to leave them if I'll add the readme files?

PR #283, opened 2020-10-22
discord submissions 2020-10-22

int-ua:
Three items by CuteCottonCandyCat


I-am-Erk:
The perspective and scaling on these is pretty far off, I think I'd like to do a cleanup before merging these ones

int-ua:
I'm merging this, we can always update later.

PR #284, opened 2020-10-22
Fix Unknown VS Unseen sprites

Fris0uman:
Unknown sprite is used when a sprite is missing, so if everything out of vision was showing as  a ? it was because lightiing_hidden and lighting_low_light were missing.

image


int-ua:
Oh, so we can have different tones of black for unlit and unseen. What do you think?

Fris0uman:
Yes we can, I have no idea how to use that so I've used the most basic set up of having stufff you can't see be black, but the artists can figure out a prettier way when they hace time to work on those

int-ua:
I mean, what will happen if you add a separate dark gray tile for `lighting_lowlight_dark`? Will it show anywhere with map memory being active?

Fris0uman:
On the screenshot `lighting_lowlight_dark` is the black beyond the vision range, and `lighting_hidden` is the black behind the shelter.
The first one is for things you can't see due to poor light, and the seconds for things you can't see because you don't have LOS.

int-ua:
I tried using 
lighting_lowlight_dark for lighting_lowlight_dark and I don't see it in the game even with map memory disabled. What am I doing wrong?

int-ua:
Almost the same image for `lighting_hidden`, just darker: lighting_hidden
Screenshots:
2020-10-24-163814_1920x1080_scrot
2020-10-24-163834_1920x1080_scrot

PR #315, opened 2020-12-02
Fix corpse sizes again

Fris0uman:
It's #306 again. I don't know why it didn't apply. It might have been overwritten by another PR but that's weird


Fris0uman:
My bad, it did work actually... ^^"

PR #341, opened 2020-12-12
Remove empty filler folders to fix #340 ASAP

### LINKED ISSUE #340: fillergiant trees are offset by two positions to the right, making a couple of them invisible in the summer
`fillergiant.png` sprites are offset by two to the right somehow. And `compose.py` always creates sheets with set width, so there is empty space to the right of the last two. If the sheet offset is changed they only become black, as we've seen previously. Maybe the problem is in the `compose.py` or even in the game.

tile_info.json:
```json
  {
    "fillergiant.png": { "sprite_width": 96, "sprite_height": 96, "sprite_offset_x": -32, "sprite_offset_y": -64, "filler": true  }
  },
```

`giant.png` looks fine, even the plums on the last row, with almost identical config:
```json
  {
    "giant.png": { "sprite_width": 96, "sprite_height": 96, "sprite_offset_x": -32, "sprite_offset_y": -64  }
  },
```

Originally reported by acepleiades
https://discord.com/channels/552510581161066497/552510581782085654/787093325551370322

int-ua:
Fixes #340 (workaround, working on the full fix for `compose.py`)

### DIFF
--- a/gfx/UltimateCataclysm/tile_info.json
+++ b/gfx/UltimateCataclysm/tile_info.json
@@ -41,12 +41,6 @@
     }
   },
   {
-    "filler.png": { "sprite_width": 32, "sprite_height": 32, "filler": true  }
-  },
-  {
-    "filler_tall.png": { "sprite_width": 32, "sprite_height": 64, "sprite_offset_y": -32, "filler": true  }
-  },
-  {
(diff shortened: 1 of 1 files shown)


int-ua:
https://github.com/CleverRaven/Cataclysm-DDA/issues/45996

PR #345, opened 2020-12-12
Add a PR template

Fris0uman:
Add A PR template to normalise all our PRs and maybe setup some automatised process to build changelogs


int-ua:
Can we have a separate simpler version for sprite-only additions? Testing section would always be the same, no need to repeat the same text, maybe no Additional context section either?

Also, this main template needs an Infrastructure category.

int-ua:
Just in case: I do not oppose this change. I do not particularly like it either as it will increase the amount of work I need to do but I can deal with it.

Fris0uman:
Additional context is optional it's just to add stuff if you it's needed.

.github/pull_request_template.md
```diff
@@ -17,17 +17,18 @@ Do not enter the square brackets [].  Tileset must be one of these:
...
 
 #### Content of the change
 
-<!-- Explain what does this pull request contains.  -->
+<!-- Explain what does this pull request contain.  -->
 
 #### Testing
 
-<!-- Describe what steps you took to test that this PR resolved the bug or added the feature, and what tests you performed to make sure it didn't cause any regressions.  Also include testing suggestions for reviewers and maintainers.  -->
+<!-- Describe what steps you took to test this PR, if applicable include screenshot of the sprites in game or at least in editor.  -->
```
int-ua, line 30: Maybe "image editor"? I was confused for a second :)

int-ua:
What if we made a short documentation file that could be linked from the template and leave only one comment to remove, with one sentence and the link?

int-ua:
I meant only comments, leaving the headings in place. And maybe not one sentence but put the whole description in one comment so it would be easier to remove in one action, not having to repeat the same thing for every heading.

Fris0uman:
Comments are not visible, so there's actually no reason to remove them

Fris0uman:
Basically if PRs could be:
```
SUMMARY: [Category] "[Briefly describe the change in these quotation marks]"

>List of added sprites

>Screenshots of those sprites
```
It would make them pretty easy to review

int-ua:
> Comments are not visible, so there's actually no reason to remove them

hmm. I guess you are technically correct. It's me who considers not removing them "untidy"/non-neat.

int-ua:
What are your thoughts on moving summary to the PR title?

Fris0uman:
Most of the times I put exactly the same thing as summary and as title so it's fine with me. But I don't see how to put that in the template

int-ua:
> Most of the times I put exactly the same thing as summary and as title so it's fine with me. But I don't see how to put that in the template

Yeah, looks like it's impossible currently: https://github.com/isaacs/github/issues/1252

int-ua:
https://github.com/Fris0uman/CDDA-Tilesets/pull/3

.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
+<!-- HOW TO USE: Under each "#### Heading" below, enter information relevant to your pull request.
+Comment blocks, surrounded with <!–– and ––>, won't be visible in the actual post.-->
+
+#### Summary
+
+SUMMARY: Category "Brief description"
```
int-ua, line 6: What if we get rid of the literal `SUMMARY:` and leave just `Category "Brief description"`?

int-ua: The fact that it's a summary is already conveyed by the section title, isn't it?

int-ua:
Oh, I executed wrong `git push` as I was trying to create a PR for your branch. I hope you are okay with these changes?

int-ua (approved):

PR #350, opened 2020-12-14
compose.py --use-all option in CI to reduce JSON burden on contributors

int-ua:
Details: https://github.com/CleverRaven/Cataclysm-DDA/pull/45950


int-ua:
Requesting reviews from all of you because it's a significant change. There will be no more warnings for unused sprites if we merge it.

Fris0uman:
Could we have a list of in game Ids that the script would check to tell us if the sprite will be used in game?

int-ua:
Neither https://github.com/CleverRaven/Cataclysm-DDA/blob/master/gfx/tile_config_template.json nor https://github.com/CleverRaven/Cataclysm-DDA/blob/master/tools/json_tools/generate_overlay_ids.py will have every possible valid id so IMO our tracker should be the verification. I'm updating it right now.

int-ua:
Yes, having some verification would've been nice. But before that we will have to fix the mismatched already discovered in the tracker and that's quite a lot.

int-ua:
Created a new issue for verification: #352

Do you agree that verification is not a blocker for this?

int-ua:
I plan to merge it tomorrow if there will be no objections. We can always revert it.

Fris0uman:
Sorry I'm late to the party ^^"
Yeah it seems ok, but we porbably need to use/pay attention to the tracker a bit more then.

PR #358, opened 2020-12-19
Bridge ramps

int-ua:
Waits for ~~https://github.com/CleverRaven/Cataclysm-DDA/pull/46139 https://github.com/CleverRaven/Cataclysm-DDA/pull/46358~~
screenshot


int-ua:
CI fail should be fixed by https://github.com/CleverRaven/Cataclysm-DDA/pull/46142

PR #388, opened 2020-12-28
Omnidirectional ramps

int-ua:
Supersedes #358 until a solution within CDDA code is implemented.

~~Waits on #395 to see if the idea of making **low** ramps sloped will work for rotations.~~

2020-12-28-125859_1920x1080_scrot
2020-12-28-125854_1920x1080_scrot
2020-12-28-125822_1920x1080_scrot
2020-12-28-125819_1920x1080_scrot


I-am-Erk:
I like the art, I would suggest breaking verisimilitude and adding a faint up or down arrow on top because it's hard to see what's going on

int-ua:
I still think that @Fris0uman idea of making the low parts rotated may work but I don't have time right now.

PR #403, opened 2021-01-03
Update t_tree_maple_spring.png

acepleiades:
I made it slightly more leaf-like


acepleiades:
Tested and it's working fine

PR #428, opened 2021-01-10
Apply pupating zombie sprites for decoys, add shady pupating zombie

acepleiades:
Applies pupating zombie sprites for pupating decoys and adds shady pupating zombie


acepleiades:
papas
It works

PR #431, opened 2021-01-11
More children zombies

acepleiades:
Adds remaining zombie children sprites and edits them


acepleiades:
children
It's working

PR #432, opened 2021-01-11
Add eyebot and removes shotgun duplicates

acepleiades:
Adds eyebot sprites and removes shotgun duplicates


acepleiades:
eyebots
It's working

PR #461, opened 2021-01-20
Add turkey and fixes screen mesh wall

acepleiades:
Adds turkey sprite and changes bg of screen mesh wall from "unconnected" floor into "center" floor (correct)


acepleiades:
turkeyey
It works

PR #467, opened 2021-01-22
Trees update

acepleiades:
Adds cherry tree, removes cherry tree filler
Adds fallen leaves to autumn trees
Updates spring maple tree


acepleiades:
fallenleaves
cherryrry
It works

PR #475, opened 2021-01-23
Bookcase rotations through multitle

acepleiades:
Adds bookcase rotations through multitiles (experimental)


acepleiades:
bookcasescasescases
Finally, it works

int-ua:
2021-01-24-104119_1920x1080_scrot
LGTM

PR #478, opened 2021-01-24
[Chibi_Ultica] Setup Symlinks for Chibi Ultica again

Fris0uman:
Setup proper symbolic links and workflow to be able to build Chibi Ultica from MSX and Ultica reliably.

I had to move around some folder in MSx so that's why it shows a lot of files modified


int-ua:
You're welcome :) It also needs to be tested by someone who did not setup Git specifically for symlinks on Windows. Others may experience links being converted after they clone repo and try making a PR.

int-ua:
If it will not work and will require specific git setup on Windows to work with repo I can offer you to create a repo at the CDDA-Tilesets org.

.github/workflows/Chibi_ci_build.yml
```diff
@@ -0,0 +1,72 @@
...
+      - 'gfx/MShockXotto\+/pngs_tiles_32x32/character/**'
+      - 'gfx/MShockXotto\+/pngs_tiles_32x32/monster/**'
+      - 'gfx/MShockXotto\+/pngs_tall_32x48/**'
+      - 'gfx/Chibi_Ultica/**'
+      - '.github/workflows/Chibi_ci_build.yml'
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'gfx/UltimateCataclysm/**'
```
int-ua, line 25: Will this action take precedence over Ultica builder?

Fris0uman: I was assuming that both action would take place, but I don't know.

Fris0uman:
image

Just did some testing and editing something in Ultica does trigger both workflow porperly

Fris0uman:
Damn it looks like it doesn actually work ><
When composing the script appears to skip or maybe not find the MSX sprites

int-ua:
> Just did some testing and editing something in Ultica does trigger both workflow porperly

What I meant that with this state anyone waiting for an Ultica Artifact will have to wait additional time for the Chibi Ultica to be built. If you agree that Ultica should take precedence I suggest trying to rename the Ultica composer workflow file to something like `0_Ultica_ci_build.yml`.

Fris0uman:
Sure I can change the order, but I'm pretty sure the two workflow run in parallele and you don't need to wait for chibi to finish to get Ultica

Fris0uman:
Both workflow are running at the same time, so if someone is only interested in Ultica this change won't affect them

PR #482, opened 2021-01-25
Display rack connections and bed edits

acepleiades:
Adds display rack connections
Edits bed frame and makeshift bed


acepleiades:
racksbed
It works

PR #493, opened 2021-01-30
Tongs

aoae23:


int-ua:
Thank you for trying, there is one issue though: the new sprite is not in a sub-directory and will be ignored in this state, please put it into a sub-directory matching the PNG size.

aoae23:
No clue how I made such a silly error, should be fixed now.

PR #496, opened 2021-01-31
[Chibi_Ultica]  Add workflow to make Chibi_Ultica release 

Fris0uman:
Add workflow to make chibi_ultica release withtout having to do it locally.

Depends on #478


Fris0uman:
Closing because we already never use the one for Ultica so this seem unecessary after all

PR #510, opened 2021-02-06
ice spike spell explosion test

int-ua:
https://github.com/CleverRaven/Cataclysm-DDA/pull/47277
out_4x


int-ua:
Mod support should be implemented with separate sheets IIRC and this is was a quick and dirty solution anyway. If anyone wants to pick it up and merge anyway I wouldn't mind.

PR #516, opened 2021-02-10
Add ammo sprites

acepleiades:
Adds ammo sprites that are a generic box with overlayed text (except for RPG rockets)


acepleiades:
ammunition
It works

It doesn't look too good when various ammo sprites are scattered like this, but it looks a lot better in a 1-tile ammo pile, which I usually tend to do. If the text is too jarring, you can store ammo in a container that is going to hide the sprite when you're not adjacent.

PR #519, opened 2021-02-11
Misc additions 2

acepleiades:
Effect icons overhaul

Update attack animation

Add item sprites for:
Leather seat
Bandana
Solder
Scarf

Replace item sprites:
Smartphone


acepleiades:
misctest
The sprites work but the effect icons appear below the character (even without clothes). As only a small part of the icons are overlayed this PR should still be merged.

PR #522, opened 2021-02-12
Bookcase fix

acepleiades:
Fixes vertical bookcase connection hiding items on the northern end


acepleiades:
bookcase fix
It's working

PR #521, opened 2021-02-12
Use ImgBot to optimise images.

UmbralReaper:
Use ImgBot to optimise image size.

**Why?**
There are three main reasons:
  - Increase download time
  - Increase game load time
  - I felt like it

If necessary I can split this into smaller PR's with the summary of what's changed.
@I-am-Erk

Total -- 28,947.94kb -> 15,511.25kb (46.42%)

PR #523, opened 2021-02-13
Use ImageMagick convert to reduce tilesheet file size

int-ua:
In addition to https://github.com/CleverRaven/Cataclysm-DDA/pull/47453 and https://github.com/CleverRaven/Cataclysm-DDA/pull/47519
See also: #521

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -49,8 +49,10 @@ jobs:
             original_size=$(stat -c%s "$file")
             imagemagick_size=$(stat -c%s "$file.tmp")
             if [ "$imagemagick_size" -lt "$original_size" ]; then
+              echo "ImageMagick shrinked $file"
               mv "$file.tmp" "$file"
             else
+              echo "ImageMagick did not shrink $file"
               rm "$file.tmp"
             fi
           done
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -45,14 +45,13 @@ jobs:
           python3 build/compose.py --use-all --obsolete-fillers gfx/UltimateCataclysm
 
           for file in gfx/UltimateCataclysm/*.png; do
-convert -type optimize -define png:compression-strategy=0 -define png:compression-level=9 -define png:exclude-chunk=EXIF,iCCP,iTXt,sRGB,tEXt,zCCP,zTXt,date "$file" "$file.tmp"
-            original_size=$(wc -c <"$file")
-            imagemagick_size=$(wc -c <"$file.tmp")
+            convert -type optimize -define png:compression-strategy=0 -define png:compression-level=9 -define png:exclude-chunk=EXIF,iCCP,iTXt,sRGB,tEXt,zCCP,zTXt,date "$file" "$file.tmp"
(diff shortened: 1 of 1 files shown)


int-ua:
TODO: use `stat` instead of `wc`

.github/workflows/ci_build.yml
```diff
@@ -43,6 +43,18 @@ jobs:
 
           cd ..
           python3 build/compose.py --use-all --obsolete-fillers gfx/UltimateCataclysm
+
+          for file in gfx/UltimateCataclysm/*.png; do
+            convert "$file" "$file.tmp"
```
UmbralReaper, line 48:
```
convert -type optimize -define png:compression-strategy=0 -define png:compression-level=9 -define png:exclude-chunk=EXIF,iCCP,iTXt,sRGB,tEXt,zCCP,zTXt,date "$file" "$file.tmp"
```
Honestly, imo, `-type optimize` isn't necessary, but it's worth having.

int-ua: Applying but don't have time to check results right now, thank you!

int-ua: Even better, please convert your comment to change suggestion (ctrl+g, `±` button in the edit interface) so that you will be properly added as a `Co-authored-by` without me needing to spending time on it :)

int-ua: зображення
Looks like the images are already pretty optimized but these options still were able to save some bytes :+1:

int-ua:
After changing filter type in https://github.com/CleverRaven/Cataclysm-DDA/pull/47519 gains from `convert` are rather small:
```
ImageMagick shrinked gfx/UltimateCataclysm/centered.png (1350 > 946)
ImageMagick shrinked gfx/UltimateCataclysm/incomplete_large.png (3918 > 3062)
ImageMagick shrinked gfx/UltimateCataclysm/incomplete_tall.png (3684 > 3010)
ImageMagick did not shrink gfx/UltimateCataclysm/fillergiant.png (19501 <= 31403)
ImageMagick did not shrink gfx/UltimateCataclysm/fillerhoder.png (44330 <= 47206)
ImageMagick did not shrink gfx/UltimateCataclysm/giant.png (75302 <= 127817)
ImageMagick did not shrink gfx/UltimateCataclysm/huge.png (16014 <= 26900)
ImageMagick did not shrink gfx/UltimateCataclysm/incomplete.png (217530 <= 282651)
ImageMagick did not shrink gfx/UltimateCataclysm/large.png (81313 <= 137885)
ImageMagick did not shrink gfx/UltimateCataclysm/normal.png (714663 <= 1103696)
ImageMagick did not shrink gfx/UltimateCataclysm/small.png (31850 <= 49680)
ImageMagick did not shrink gfx/UltimateCataclysm/tall.png (282637 <= 412249)
```

There is also this but fallback.png optimization is included in #530
```
ImageMagick shrinked gfx/UltimateCataclysm/fallback.png (316141 > 191846)
```

So now I'm not sure if it's worth it.

int-ua:
I'll close this for now, but I don't mind reopening it if someone else thinks all the extra CPU time on conversion is worth saving 2KiB.

PR #553, opened 2021-03-05
[MSX] Port rotary clothline from Ultica

Fris0uman:
#### Summary
MSX "Port rotary clothline from Ultica"

#### Content of the change

rotary clothline

#### Testing

Untitled


Fris0uman:
@int-ua Apprently when writing the json compose puts the sprite in the normal tilesheet instead of the Tall and I don't know why

PR #595, opened 2021-04-15
[MD] Mushroom dream rebase

AlbertTheTerrible:
#### Summary
MD "More tiles"

#### Content of the change
Rebase to fix conflicts


gfx/Mushroom-Dream/pngs_titan_160x128/mon_shoggoth.json
```diff
@@ -0,0 +1,6 @@
+
+{
+    "id": "mon_shoggoth",
+    "fg": [
+        { "weight": 1, "sprite": "mon_shoggoth" }
```
int-ua, line 5: just `"fg": "mon_shoggoth"` if you insist on having JSON. Add the array with weights only when you will have more than one element there

int-ua:
#596

PR #596, opened 2021-04-15
MD "trying to fix this"

AlbertTheTerrible:
#### Summary
MD "trying to fix this"


gfx/Mushroom-Dream/pngs_titan_160x128/mon_shoggoth.json
```diff
@@ -0,0 +1,4 @@
+{
+    "id": "mon_shoggoth",
```
Fris0uman, line 2: this json is unnecessary since the sprite file is already called `mon_shoggoth`

gfx/Mushroom-Dream/pngs_tall_32x48/player_male.txt
```diff
@@ -0,0 +1,10 @@
+{
+    "id": "player_male",
```
Fris0uman, line 2:
```suggestion
    "id": [ "player_male", "player_female" ]
```

Fris0uman (commented):
I think you need to flip the shoggoth so that it looks to the right, otherwise they will always look like they're walking backward in game

gfx/Mushroom-Dream/pngs_tall_32x48/player_male.txt
AlbertTheTerrible: i incorporated this, on the other PR, i'm gonna close this one because i'm afraid it'll cause trouble again 🎃

PR #652, opened 2021-06-12
Log wall fixes

acepleiades:
#### Summary
Ultica "log wall fix"

#### Content of the change
Removes transparency from log wall tiles


acepleiades:
walls
In the end I just ended up filling the empty pixels but it works

PR #653, opened 2021-06-16
Add deck chairs by hoasd

acepleiades:
#### Summary
Ultica "deck chairs"

#### Content of the change
Add deck chairs by hoasd


acepleiades:
deckchairs
It works, but I couldn't get the variations to work. I'm not sure if that's even supported for furniture

PR #661, opened 2021-06-21
Add filler sprites for rams and seed drill

Fris0uman:
#### Summary
Ultica "Add filler sprites for rams and seed drill"

#### Content of the change

filler for rams and seed drill

#### Testing

Untitled


I-am-Erk:
I should get around to drawing those...

PR #689, opened 2021-07-01
A lot of tree edits

acepleiades:
#### Summary
Ultica "A lot of tree edits"

#### Content of the change
Edits every tree sprite


acepleiades:
fixed trees
It works

PR #705, opened 2021-07-06
Overmap silos

acepleiades:
#### Summary
Ultica "Overmap silos"

#### Content of the change
Adds overmap silos sprites


int-ua (approved):

acepleiades:
wrong silos
I can't get the right rotation to work, this one is supposed to be silos_NS but is silos_EW instead

acepleiades:
silos ok
I got them to rotate correctly

PR #706, opened 2021-07-06
[RetroDays] Add soil tile

anothersimulacrum:
#### Summary
RetroDays "Add soil tile"

#### Content of the change
I'm adding soil tiles to DDA, so I'm adding some sprites first.

#### Testing
image
image

#### Additional information
There's no way for anyone else to test this, at the moment, sorry.


Fris0uman:
You don't need the json, since the sprite is already called t_soil

anothersimulacrum:
When I do that, I get `Warning: sprite filename t_soil was not used in any main tile_config.json entries`

Fris0uman:
You need to use the `--use-all` option of the compose script

PR #707, opened 2021-07-06
Ultica "alternators, new fire, polisher, barometer, note_!_*"

int-ua:
#### Summary
Ultica "alternators, new fire, polisher, barometer, note_!_*"

#### Content of the change
Screenshot_20210710_155709

#### Additional information

f_bigmirror should be made out of metal ideally


int-ua:
@I-am-Erk do you think the new fire is better?

PR #710, opened 2021-07-07
Overmap bridges

acepleiades:
#### Summary
Ultica "Overmap bridges"

#### Content of the change
Adds Overmap bridges sprites


acepleiades:
bridge
bridge2
It works

PR #711, opened 2021-07-07
Overmap multitile river banks and open air

acepleiades:
#### Summary
Ultica "Overmap multitile river banks and open air"

#### Content of the change
Adds open air sprite and makes river banks multitile


acepleiades:
bizarre bridge
I cannot get the edges to work correctly

acepleiades:
good river
It works

gfx/UltimateCataclysm/pngs_normal_32x32/overmap/river/river.json
```diff
@@ -0,0 +1,71 @@
+[{
+  "id": "river_nw",
+  "fg": ["river_nw"],
```
int-ua, line 3:
```suggestion
  "fg": "river_nw",
```
You should be able to omit arrays if there is only element. Please try.

PR #712, opened 2021-07-07
Overmap barn

acepleiades:
#### Summary
Ultica "Overmap barn"

#### Content of the change
Adds Overmap barn sprites


acepleiades:
barnss
It works

PR #713, opened 2021-07-08
Overmap pasture

acepleiades:
#### Summary
Ultica "Overmap pasture"

#### Content of the change
Adds Overmap pasture sprites, but the IDs to which the sprite must be applied need to be discovered. Currently only working for "2farm"


acepleiades:
2farm overmap
It works

PR #714, opened 2021-07-08
Overmap cave

acepleiades:
#### Summary
Ultica "Overmap cave"

#### Content of the change
Adds Overmap cave sprite


acepleiades:
caveee
It works

PR #720, opened 2021-07-09
Remembered vehicle sprite

acepleiades:
#### Summary
Ultica "Remembered vehicle sprite"

#### Content of the change
Adds an icon called overmap_remembered_vehicle that will be used for remembered vehicle in the overmap
Edit: also adds horde icons


acepleiades:
icons
It works

acepleiades:
https://github.com/CleverRaven/Cataclysm-DDA/pull/49814
The problem with the horde icons don't seem to be a problem with the tileset, so this PR can be merged

PR #725, opened 2021-07-11
Ultica "leather patch, bread-related food, pizza"

int-ua:
#### Summary
Ultica "leather patch, bread-related food, pizza"

#### Content of the change
leather_4x

pizza

Some bread-related foods derived from https://opengameart.org/content/lpc-food
breads


I-am-Erk:
That's a bit too colour smoothed I think

int-ua:
which one is better?
out_4x

acepleiades:
It works

PR #730, opened 2021-07-14
Overmap small cemetery

acepleiades:
#### Summary
Ultica "Overmap small cemetery"

#### Content of the change
Adds Overmap small cemetery sprite


acepleiades:
cemetery
It works, I also made headstones use the gravestone sprite

PR #733, opened 2021-07-14
Ultica overmap ranch camp json

acepleiades:
#### Summary
Ultica "Ultica overmap ranch camp json"

#### Content of the change
Adds overmap ranch camp json


acepleiades:
ranch camp
Good enough for now, pond and silo ideally need their own sprite and garage is still missing, but better than before

PR #737, opened 2021-07-15
Ultica radio tower

acepleiades:
#### Summary
Ultica "Ultica radio tower"

#### Content of the change
Adds Ultica radio tower sprite


acepleiades:
towers
It works

int-ua:
Should it also be used for overmap?

acepleiades:
> Should it also be used for overmap?

I think overmap needs it's own sprite, like an entire radio tower.

PR #741, opened 2021-07-15
Ultica rope and pulley

acepleiades:
#### Summary
Ultica "Ultica rope and pulley"

#### Content of the change
Adds rope and pulley sprite


acepleiades:
ropepulley
It works

int-ua:
Are the two files identical? Why not use one with JSON to reduce tilesheet size?

acepleiades:
> Are the two files identical? Why not use one with JSON to reduce tilesheet size?

Done

PR #742, opened 2021-07-16
Ultica json for "special" otr

acepleiades:
#### Summary
Ultica "json for "special" otr"

#### Content of the change
Adds json for "special" overmap terrain, which seems to be used for isherwood quest. This is still an incomplete task because some terrain tiles are still missing altogether, but it's an improvement


acepleiades:
special
It works

PR #743, opened 2021-07-16
Ultica adapted overlays: bows and crossbows

acepleiades:
#### Summary
Ultica "adapted overlays: bows and crossbows"

#### Content of the change
Adapts overlays for wielded bows and crossbows into item sprites


acepleiades:
range
It works

PR #749, opened 2021-07-19
Ultica mobile sprites by erk and recolors

acepleiades:
#### Summary
Ultica "mobile sprites by erk and recolors"

#### Content of the change
Adds sprites by erk made on mobile and recolors blob monsters and boulders


acepleiades:
newthings
It works

PR #785, opened 2021-08-05
Ultica "32x36 sheet, phase 1: overlays"

int-ua:
#### Summary
Ultica "32x36 sheet, phase 1: overlays"

#### Content of the change
Moved 99% of all sprites in tall/overlays to a new 32x36 sheet.

#### Testing
Tested it in the game and it's working, no time to make screenshots


int-ua:
There is somehow a default skin 32px above the character now but I'll figure it out tomorrow

int-ua:
I've put all affected images in two directories: original and edited PNGs. Then I've made 32x36 back into 32x64 with
```sh
mogrify -gravity north -background none -extent 32x40 *.png
mogrify -gravity south -background none -extent 32x64 *.png
```
After that I"ve made a sheet out of each directory
```sh
montage `ls -1 *.png | sort -V` -geometry 32x64+0+0 -tile 16x85 -background none output.png
```
And I don't see a single pixel of difference between them while using

int-ua:
image-difference

PR #795, opened 2021-08-09
Msx temporary vehicle fix

acepleiades:
#### Summary
MSX "temporary vehicle fix"

#### Content of the change
Adds a temporary vehicle fix json for MSX for when https://github.com/CleverRaven/Cataclysm-DDA/pull/50632 gets merged


acepleiades:
msx fix
It appears to be working

PR #797, opened 2021-08-10
BLB vehicle temporary fix

acepleiades:
#### Summary
BLB "vehicle temporary fix"

#### Content of the change
Temporary vehicle sprites fix for when https://github.com/CleverRaven/Cataclysm-DDA/pull/50632 gets merged


acepleiades:
blb fix
Appears to be good

PR #798, opened 2021-08-10
Retrodays vehicle fix

acepleiades:
#### Summary
RetroDays "vehicle fix"

#### Content of the change
Adds sprites to new vehicle part shapes in https://github.com/CleverRaven/Cataclysm-DDA/pull/50632


acepleiades:
retrodays fix
It works

anothersimulacrum:
The van in the top left should have three doors in a row on the left, not a door, then a board, then a door.

acepleiades:
> The van in the top left should have three doors in a row on the left, not a door, then a board, then a door (unless you changed the vehicle layout).

I did change the layout, I am fully aware that some may find this a selfish decision but it is part of the "deliberate choices to make certain vehicles look better with the new sprites for UltiCa (with care to make them look right in other tilesets)"
Why did it change? Because it would be very difficult making a connection for a board-aligned sliding door and the cab in the front of the hippie van
ultica hippie van
If you feel strongly about having the door of the hippie van in the previous location, let me know if you want to me come up with a better solution. But I took care to make things look right in other tilesets. I really care.

anothersimulacrum:
I don't care, just wanted to make sure there were no errors.

PR #799, opened 2021-08-10
Neodays vehicle fix

acepleiades:
#### Summary
NeoDays "vehicle fix"

#### Content of the change
Adds sprites for the new vehicle parts in https://github.com/CleverRaven/Cataclysm-DDA/pull/50632


acepleiades:
neodays fix
My parts now have sprites, but Neodays seems to have some missing sprites for parts that have nothing to do with my changes. So it appears to be working

PR #802, opened 2021-08-10
[MSX] Some furniture rotations

Fris0uman:
#### Summary
MSX "Some furniture rotations"

#### Content of the change

filing cabinet rotations, vending machine rotations

#### Testing

Untitled2
Untitled

image

image


EmpressWynnter:
The vending machine should probably be a little wider in the side view, and perhaps would look nicer if the letters were colored in on "soda" with some bright colours. 😶

PR #817, opened 2021-08-17
Ultica more effect overlays

acepleiades:
#### Summary
Ultica "more effect overlays"

#### Content of the change
Adds more effect overlays

#### Testing
debugjon
Most of them work. Evil and wet didn't work. But I also saw that they didn't work in UDP, which had sprites for them as well


actual-nh:
Thanks!

PR #837, opened 2021-08-23
Yellow overlay for prone movement mode

actual-nh:
#### Summary
Ultica "Yellow overlay for prone movement mode"

#### Content of the change

This is an alternate overlay for prone movement mode. It is distinct from the old icon for the crouched movement mode in being yellow; this makes it easier to distinguish from the crouched movement mode icon for players used to that one, conveys that the prone movement mode is very slow, and does not look too much like red HP, etc indicators.

#### Testing

I actually use curses, not tiles, so have not had an opportunity to test it. However, it is a direct transformation (using ImageMagick) from the original (purple) overlay used for the crouched movement mode. Here is an enlarged image:
overlay_mf_crouch max3 contrast-fx r400

#### Additional information

Ping: @acepleiades?


acepleiades:
Thank you for getting into this, this PR should be merged.

PR #843, opened 2021-08-24
Add a script to draw black outline around sprites

Fris0uman:
#### Summary
Infrastructure "Add a script to draw black outline around sprites"

#### Content of the change

Add a script to draw black outline around sprites. Very usefull to mass convert Ultica's sprite to MSX.

#### Testing
see #844 gun sprites

### DIFF
--- a/tools/add_outline.py
+++ b/tools/add_outline.py
@@ -1,48 +1,61 @@
-# -*- coding: utf-8 -*-
+#!/usr/bin/env python3
+"""
+Add black outline to all sprites in a folder
+"""
 import os
 import glob
 import argparse
 import numpy as np
 from PIL import Image
 
-BLANK = np.array([0,0,0,0])
-BLACK = np.array([0,0,0,255])
+BLANK = np.array([0, 0, 0, 0])
+BLACK = np.array([0, 0, 0, 255])
 
-def is_outline(table: np.array, col: int, row: int):
-    
-    up = table[row -1,col,:]
-    down = table[row +1,col,:]
-    left = table[row,col-1,:]
-    right = table[row,col+1,:]
-    
-    for side in [left,right,up,down]:
+
+def is_outline(table: np.array, col: int, row: int) -> bool:
+    """
+    Determine if the point belongs to the outline
+    """
+    up = table[row - 1, col, :]
+    down = table[row + 1, col, :]
+    left = table[row, col - 1, :]
+    right = table[row, col + 1, :]
+
+    for side in [left, right, up, down]:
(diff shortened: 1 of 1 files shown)


tools/add_outline.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Add black outline around a sprite")
+    parser.add_argument(
+        "folder",
+        help="path to the folder of images to edit")
+    main(parser.parse_args())
+
+
```
int-ua, line 48:
```suggestion
#!/usr/bin/env python3
"""
Add black outline to all sprites in a folder
"""
import os
import glob
import argparse
import numpy as np
from PIL import Image

BLANK = np.array([0, 0, 0, 0])
BLACK = np.array([0, 0, 0, 255])


def is_outline(table: np.array, col: int, row: int) -> bool:
    """
    Determine if the point belongs to the outline
    """
    up = table[row - 1, col, :]
    down = table[row + 1, col, :]
    left = table[row, col - 1, :]
    right = table[row, col + 1, :]

    for side in [left, right, up, down]:
        if (side != BLACK).any() & (side != BLANK).any():
            return True

    return False


def draw_outline(data: np.array) -> None:
    """
    Add outline to the pixels array in-place
    """
    for i in range(1, 30):
        for j in range(1, 30):
            if (data[j, i, :] == BLANK).any():
                if is_outline(data, i, j):
                    data[j, i, :] = BLACK


def main(args):
    """
    Add outlines to all *.png images in the args.folder
    """
    for filename in glob.glob(os.path.join(args.folder, '*.png')):
        with Image.open(
                os.path.join(os.getcwd(), filename)).convert('RGBA') as img:
            data = np.array(img)
            draw_outline(data)
            Image.fromarray(data, mode='RGBA').save(filename, 'PNG')


if __name__ == "__main__":
    parser = argparse.ArgumentParser(
        description="Add black outline around a sprite")
    parser.add_argument(
        "folder",
        help="path to the folder of images to edit")
    main(parser.parse_args())

```

tools/add_outline.py
```diff
@@ -0,0 +1,61 @@
...
+    Add outline to the pixels array in-place
+    """
+    for i in range(1, 30):
+        for j in range(1, 30):
+            if (data[j, i, :] == BLANK).any():
+                if is_outline(data, i, j):
+                    data[j, i, :] = BLACK
+
+
+def main(args):
```
int-ua, line 42:
```suggestion
def main(args: argparse.Namespace) -> None:
```
type hinting the main

PR #845, opened 2021-08-25
Ultica farm animals

acepleiades:
#### Summary
Ultica "farm animals"

#### Content of the change
Adds cow, chicken, horse

#### Testing
animals animals
cowboy
ridden cow


acepleiades:
I made a new tilesheet for ridden sprites that is offset downward by 5 pixels. However, the 5 pixels that go below the tile that the sprite is in are cut off. They don't appear. It's a problem with the graphical engine (not tileset side).

Fris0uman:
I think riding overlays probably need to use redrawn scaled down sprites

PR #853, opened 2021-08-27
Ultica "fungal shrub, mon_dispatch_*"

int-ua:
#### Summary
Ultica "fungal shrub, mon_dispatch_*"

#### Content of the change
t_shrub_fungal1_4x
out_4x


acepleiades:
ttest
The fungal shrub does not have a background

int-ua:
Ok, dispatch robots will get tank treads in a later update, merging this as-is.

PR #859, opened 2021-08-30
[MSX] New stock pot sprite

Gules:
#### Summary
MSX "New stock pot sprite"

#### Content of the change

Added a sprite for the Stock Pot and also added sprite and JSON for it to be wielded.

#### Testing
stock_pot_capture


Fris0uman:
That's not actually an issue, but even the wielded sprite didn't need a json file since it had the correct name.

PR #867, opened 2021-08-30
Power armour!

I-am-Erk:
#### Summary
Ultica "Adds power armour and robofac prototype"

#### Content of the change
Adds worn power armour and helmets of all varieties.

Also adds a robofac prototype robot, a collaboration between me ant @int-ua 

Power armours: top male, bottom female; left light, right heavy
image

Robofac prototype
image


int-ua:
That robot is a derivative of https://www.blendswap.com/blend/28467 I would like to add a `credits.txt` with the link, would you mind?

I-am-Erk:
Please do

PR #875, opened 2021-08-31
Test combining Neo and Retro days

Fris0uman:
#### Summary
Infrastructure "Test combining Neo and Retro days"

#### Content of the change

Split Retrodays content into several sheets:
 - bigMonsters
 - nativeTransparency: sheet with transparent sprites even in Retrodays
- normal
- overmap

Remove those sprites from Neodays
Add symbolic links in Neodays that point to those tilesheets
All sprite have transparent brackground by default, the black background should be added after composing for the bigMonsters and normal tilesheets in Retrodays


Fris0uman:
I don't really want to solve the conflicts, it would just be easier to redo it if we decide that's what we want. This was mostly a proof of concept

PR #894, opened 2021-09-05
[Ultica] pool cue

Entity-y:
#### Summary
Ultica "pool cue"

#### Content of the change
adds pool cue item and wielded overlay

#### Testing
Using both Pixel Studio and GIMP, verified that the overlay of wielded pool cue lined up properly with male and female sprites in a 32x36 template and then saved position. Then adjusted and saved position of the pool cue item to fit on the surface of a counter in a 32x32 template.

Proof of concept
Screenshot_20210905-073331_Pixel Studio


int-ua:
Thank you! Ultica uses `compose.py --use-all` feature and the files can just be named directly with IDs if these IDs are used just one time in JSONs, like in this case. Would you mind removing json files and just naming the files `overlay_female_wielded_pool_cue.png`, `overlay_male_wielded_pool_cue.png` and `pool_cue.png`?

int-ua:
Almost, now just rename `pool_cue_f.png` and `pool_cue_m.png` to `overlay_female_wielded_pool_cue.png` and `overlay_male_wielded_pool_cue.png`

Entity-y:
This is a nightmare trying to rename on mobile

int-ua:
> This is a nightmare trying to rename on mobile

Haven't tried it yet :) Is different from https://docs.github.com/en/github/managing-files-in-a-repository/managing-files-on-github/renaming-a-file ? Anyway, can you bring back the wielded versions?

Entity-y:
the problem is that you need to use terminal to rename png files and I don't think that's an option on mobile

int-ua (approved):

PR #902, opened 2021-09-06
[Ultica] add robin

Entity-y:
#### Summary
Ultica "Robin"

#### Content of the change
New sprite for bird

#### Testing
image


acepleiades:
Another great contribution!
I really like the sprite but it requires a shadow. You can add the shadow by adding a json file. Let's name it mon_robin.json
In the mon_robin.json, we need to write the following:

```
{
    "id": "mon_robin",
    "fg": "mon_robin",
    "bg": "shadow_32x32_small"
}
```

Then it will have a shadow in game.

PR #922, opened 2021-09-08
Begin bringing HitButton back.

I-am-Erk:
#### Summary
Hitbutton "Some initial recolour work on hitbutton"

#### Content of the change
I'd like hitbutton back in mainline. I think it's one of our best tilesets, but the strict EGA-style colours are a bit hard on the eyes. This will bring it back with more colours and less hard black desaturation, but the same basic look.


I-am-Erk:
I think I am giong to try something different with black backroudns.

PR #923, opened 2021-09-08
begin surveyors map repo and add gunstores

curstwist:
#### Summary
none

#### Content of the change
This begins adding the surveyor's map overmap sprite set that I'm creating.  It adds folders and the gun stores with json.

#### Testing
I haven't tested in game yet, I wanted to begin adding the sprites so they don't pile up on my local drive.  

#### Additional information
Ideally I'd like this to become a separate overmap tile set (once supported by CDDA).  Other tile sets are welcome to incorporate them as fall backs, or, adjust them for their own purposes.

These are simple 32x32 flat sprites that represent the mapgen they are referencing.  It shows door, window, stair and gutter placement as well as some outdoor terrains like sidewalks and pavement.  Each sprite has a core 24x24 pixel image that replicates the mapgen.  The remaining rows are for procgen sidewalk and filled to represent the land use code associated with the mapgen.  Currently, it is blue for commercial and green for residential.

I'm largely staying with the known ascii colors for the building fill colors to help players identify with something familiar.  I am adjusting some colors to better match their "theme" (for example hunting and camping stores will get the same color as sports stores).

I need to figure out if I need to do additional pngs for the rotations.


int-ua:
(sadly looks at https://github.com/CDDA-Tilesets/ )

PR #928, opened 2021-09-09
Give filter masks a non-overlay sprite.

changemewtf:
#### Summary
Ultica "Filter mask item uses dust mask sprite"

#### Content of the change
JSON-only change to the sprite definition for dust mask, turning it into an array with an additional value.

Hopefully a trivial patch to add a sprite for filter masks and get started contributing.

This won't be very consistent with the overlay appearance, but it's
better than nothing!

#### Additional information
Great project, excited to contribute more!

### DIFF
--- a/gfx/UltimateCataclysm/pngs_normal_32x32/items/clothes/clothes.json
+++ b/gfx/UltimateCataclysm/pngs_normal_32x32/items/clothes/clothes.json
@@ -140,7 +140,7 @@
     "bg": ""
   },
   {
-    "id": "mask_dust",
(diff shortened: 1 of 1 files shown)


int-ua:
You didn't set up a feature branch and you may have trouble syncing you master with upstream. But the change itself is good, merging.

acepleiades:
This should have been done on the CDDA repo, adding a looks_like relation to the items. Like this:
https://github.com/CleverRaven/Cataclysm-DDA/pull/45953 
When someone does a sprite for the filter mask this is going to cause a headache until they figure where the sprite for the filter mask is defined.

PR #931, opened 2021-09-09
[Ultica] variants added to sludge (again)

vetall812:
#### Summary
Ultica "added variants to fd_sludge"

#### Content of the change
added two more variants to fd_sludge multitile, changed .json

#### Testing
image


acepleiades:
We need to figure out where a tab is because intua added a check that fails if there are tabs in the jsons.

PR #935, opened 2021-09-09
Add shared IDs for similar mutation overlays.

xerenogan:
#### Summary
Ultica "Add shared IDs for similar mutation overlays.(Spider)"

#### Content of the change
This adds some of the IDs for the spider skins and limbs that don't have a sprite to the existing ones to the ID lists for skins and arachnid limbs. They are very similar anyway and this way they have an appropriate looking overlay until a better one is drawn.
This way, there aren't any jarring skin changes when mutating to a missing one and seeing your character with their original skin all of the sudden.

#### Testing
Composed the tileset and overwrote ultica on a clean copy. It seems to work.

### DIFF
--- a/gfx/UltimateCataclysm/pngs_human_body_32x36/overlay/mutations/arachnid_arms/arachnid_arms.json
+++ b/gfx/UltimateCataclysm/pngs_human_body_32x36/overlay/mutations/arachnid_arms/arachnid_arms.json
@@ -1,11 +1,11 @@
 [
   {
-    "id": "overlay_female_mutation_ARACHNID_ARMS",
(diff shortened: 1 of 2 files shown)


I-am-Erk:
I'm generally against doing this on the tileset end. I don't know if mutations support looks_like, but this would be better handled by adding looks_like on the CDDA end. If we're not going to do that I think I'd prefer even duplicating the sprite and putting the duplicated sprite on a filler sheet than doing it this way, just because that makes it clear in our documentation that this sprite hasn't been drawn yet. Doing it this way will make it very hard to remember that we still need to draw the missing sprites.

xerenogan:
Ah, alrighty then. I'll close this.

PR #937, opened 2021-09-10
Start on converting files to 0-alpha gray

actual-nh:
#### Summary
Ultica "Start on converting files to 0-alpha gray"

#### Content of the change

Some sprite files have something other than black (or gray) in places where they have an alpha of 0; this is not seen, but can mess one up if either fiddling with alpha compositing or trying to take a look at the file's color distribution statistics. For the latter, it is preferable if all 0-alpha are (127,127,127,0), as in gray, which is how I have modified the files. If preferred, I can, however, upload ones that are 0-alpha black, which is currently true for the (non-problematic) Ultica sprites that I have examined. This PR changes two files, fungaloid.png and mon_zombie_4.png, which currently look as follows if one turns off the alpha (enlarged x4):

fungaloid noalpha 4x
mon_zombie_4 new noalpha 4x

(The mon_zombie_4 case is rather obvious. For the fungaloid, note the thickened back leg with a piece hanging off the side, and possibly other differences.)

Further such can be done if there is interest, although it may take me a bit to clear enough space on my hard drive and write a script (using ImageMagick). How many such files there are is a good question; note that I found these on looking at a total of 4 files (mon_zombie_3, mon_zombie_4, fd_haze, and fungaloid).

#### Additional information

See the dev discord discussion at <https://discord.com/channels/598523535169945603/598614717799596055/882478991435530241> for some further information.


actual-nh:
Converting to draft for the moment - will upload a 32x48 version of mon_zombie_4.

PR #489, opened 2021-01-29
[Ultica] Corpses and some misc items

MrAsbestos:
Adds the zombie soldier and headless zombie corpse with it's wielded overlays.

Also adds the correct sprites for the baseball and hockey puck and adds the MRE.

PR #956, opened 2021-09-11
Implement CDDA's JSON validator

casswedson:
#### Summary
None

#### Content of the change
A ~hacky~ version of cdda's json validator, in this case a compiled binary ready to be used in a json style check.
"hacky" 'cause keeping compiled binaries around is not really an elegant solution to anything, however this still is a simple solution to the need for a portable json validator.
- Is there a need for a json validator test?
The formatter is linked in the discord as a json cleaning tool. <https://discord.com/channels/552510581161066497/552510581782085654/884897280921055322>
See the other json/python style check
- Big upside to using this instead of other CLI tools <https://dev.narc.ro/cataclysm/format.html> is available to everyone.

Also include a bit of documentation on the formatter and how to use it, a bash script to automate linting and lastly format everything.

#### Testing
CI

#### Additional information
Me and @int-ua  talking about this <https://discord.com/channels/552510581161066497/552544143499460623/876884033173327922>


acepleiades:
> Is there a need for a json validator test?
> The formatter is linked in the discord as a json cleaning tool.

It was only linked because intua added the tab check and vetal's PR failed because of it.
In practical terms I can't help but only see this as hurdle for new contributors like vetal getting PR's done, as Notepad++ adds tabs by default.
Doesn't the json become one json after it's composed? What are the benefits of adding these checks? No one ever had a problem referencing existing jsons before.
I'd like a better answer to that question.

casswedson:
I don't really have many good arguments for why this is useful, other that my initial reason to start with this, int_ua's frustrations? (this is an example https://github.com/I-am-Erk/CDDA-Tilesets/pull/489#discussion_r566988558), I had this around and asked if this was OK for me to do, int_ua said :
image
This is a hurdle? yes it is, I understand your point, I just don't bring this here with bad intentions.

Fris0uman:
We rarely go back to editing already made json in the tileset repo, so I'm not sure enforcing it to be all linted the same is necessary.

int-ua:
https://www.youtube.com/watch?v=tnzz-eFmKaw

I'll be back later, have some deadlines to meet.

casswedson:
Since there is not much need for this apart from reducing OCD levels a lil bit, I am closing this, proof of concept completed.

PR #958, opened 2021-09-11
[SurveyorsMap] Create fallback.png

curstwist:
#### Summary
Smap "Add fallback.png"


curstwist:
This isn't correct, closing.

PR #959, opened 2021-09-11
[MSX] Use symlink to use SurveyorMap as filler overmap

Fris0uman:
#### Summary
MSX "Use symlink to use SurveyorMap as filler overmap"

#### Content of the change

Use symlink to use SurveyorMap as filler overmap


Fris0uman:
24x24 sprite might not look so good in a 32x32 grid

PR #1001, opened 2021-09-16
Feline mutation sprites

KittyTac:
#### Summary
Ultica "I added all the missing mutation sprites for the feline path"

#### Content of the change
See Summary, I didn't add the retractable claws yet, I think they'd be too small to be seen even when extended, looking at an actual cat's claws.

#### Testing
The mockup I made by pasting the tiles together:
image

Will edit this as soon as I download the artifact after the test I think, but as I made this with the same method as the lupine ones, should work fine.

#### Additional information
N/A


KittyTac:
I forgot to actually add the sleek fur sprite. Closing and reopening.

PR #1014, opened 2021-09-18
Add quills mutation tile

KittyTac:
#### Summary
Ultica "I added a tile for the quills mutation"

#### Content of the change
See summary.

#### Testing
image

#### Additional information
Sorry if my PRs are too spammy, if you tell me I will consolidate them into fewer big ones.


acepleiades:
Copying and pasting the discussion from discord

Feedback from me
> Well I think that effect could be achieved by using different shades for the quills
> Closer it is to the viewer, lighter shade
> Then also give them some color further away from their start
> I did draw many many animals recently and I know it's possible...
> But I also understand that it would be a difficult effect to achieve

Feedback from erk
> I think there does need to be something denser closer to the body, it would look a bit like fur... And then those long quills as you've done could be longer and lighter coloured to stand out

KittyTac:
Working on the improved sprite.

Fris0uman:
Is this ready?

Fris0uman:
I'm merging it, we can always revert it if we decide we're not happy about it

PR #1049, opened 2021-09-28
Ultica vehicle boards new perspective

acepleiades:
#### Summary
Ultica "vehicle boards new perspective"

#### Content of the change
Ultica vehicle boards new perspective


acepleiades:
firetruck

PR #1098, opened 2021-11-05
SmashButton: the new HitButton iso.

I-am-Erk:
#### Summary
HitButton "Create SmashButton Iso, a HitButton revitalizing project"

#  Content of the change
The goal here is to keep the *feel* of HitButton, but to clean up and upgrade a lot of the visuals. The resulting tileset should have the same "4 bit EGA graphics feel" and simple art style, and be isometric, but should be cleaner and easier on the eyes.

The original HitButton was not actual 4bit and contains around 111 colours. In a similar vein to both original hitbutton and UltiCa, SmashButton uses a full colour palette but aims to avoid more than 2 colours plus black outlines in most sprites. I have added outlines to furniture and terrain to sharpen boundaries and create a clearer visual read. At the same time, I have reduced the amount of black in floor/background tiles, to reduce contrast and eyestrain, and I have generally redrawn floors to have more detail but be darker in colour overall. It is still important, especially with floors, to keep the sense of a pervasive black background - this is vital to maintaining the art style.

I have added a "highlight" to monsters and characters, with different shapes signifying different types of monster (circle for players and npcs, hex for zombies, triangle for animals, square for robots, for example). The colour of the shape is less important but can contain a bit of clues to what powers they have, and also just helps to differentiate a crowd. Later this will allow characters and monsters to have a bit more detail without getting hard to read at a glance.

I've added many new tiles and a lot of tile variation to create a new feel. I have redrawn the basic walls to get a less obviously tiled look and feel a bit more natural; this will extend to all walls soon. The character sprite is likely to be replaced but I will try to avoid that in this initial PR. I've redone doors to show the swung-open door frame, and I hope to carry this on to broken doors etc. as well before I finish. Many furnitures have been updated with a bit more detail when I went in to outline them.

# Testing
## The Original HitButton Iso
image
image
image

## SmashButton Iso
image
image
image

# Additional information


Code-Moss:
damn, this looks sick asf, hope to see it take off ^^

PR #1101, opened 2021-11-09
[Ultica]Updated Blood For The New Numbering System

bombasticSlacks:
#### Summary
Ultica "Updated multi intensity blood for new numbering system"

#### Content of the change
just fixes the naming of the files for blood for https://github.com/CleverRaven/Cataclysm-DDA/pull/52727
shouldn't be merged until that goes in. Which it has now!

#### Testing

working on it


bombasticSlacks:
@Fris0uman I know you just updated the sprite set but with [CleverRaven/Cataclysm-DDA#52727 ](https://github.com/CleverRaven/Cataclysm-DDA/pull/52727) now merged blood is going to be broken and behaving weird in Ultica without this change.

PR #1107, opened 2021-11-12
[Ultica] add layering example

bombasticSlacks:
#### Summary
Ultica "Added a desk example for layering item sprites"

#### Content of the change
Adds layer sprites (implemented in: https://github.com/CleverRaven/Cataclysm-DDA/pull/52632) for desks so people can get an idea of how it works. Also added a basic sheet without much on it for Altica and Chibi Ultica, however I'm not sure one where they get all their sized sheets from so I didn't add everything to them. 

#### Testing

Ultica:

image

image


Fris0uman:
Don't add json directly to Altica and Chibi Ultica, they should copy Ultica json through symbolic links

bombasticSlacks:
The issue is that some layering sprites are tall 32x64 and it seems altica and chibi ultica don't copy ultica for that but instead MSX and it's own alt folder. So copying the json for layering from Ultica breaks them.

I'll just remove the changes to altica and chibi since I don't fully understand the infrastructure and leave this just for Ultica.

gfx/UltimateCataclysm/pngs_tall_32x64/layers/desk_laptop.json
```diff
@@ -0,0 +1,5 @@
+{
+  "id": "desk_laptop",
+  "fg": "desk_laptop",
```
Fris0uman, line 3: You don't need this json if the id is the name of the sprite, or am I missing something?

bombasticSlacks: No you don't thanks I'll remove it

PR #1116, opened 2021-11-26
Add many new exoplanet terrain tiles to Retrodays/Neodays

John-Candlebury:
#### Summary
NeoDays "Add many new exoplanet terrain tiles to Retrodays/Neodays"

#### Testing
Heres some of them:

image

image


Fris0uman:
It looks like some of the Retrodays sprite have transparent background instead of black

PR #1120, opened 2021-12-01
[Ultica] Added flowers

nheve:
#### Summary
Ultica "Added flowers"

#### Content of the change
cattail rhizome, lotus, lotus flower, dandelion, spurge flowers

Added cattail_rhizome.png
Added f_lotus.png
Added lotus.png
Added raw_dandelion.png
Added spurge.png

#### Testing
cattails_screen1
flowers_screen

#### Additional information
...


int-ua:
You don't need to add empty values like `"bg": ""` and you don't even need tile entries where there is only `id` and `fg` that are the same if the fg is not used anywhere else. It's not a big deal but why add files that are redundant? LGTM otherwise

nheve:
Sorry for including redundant .json files, already deleted them. Thank you for pointing that out int-ua - i will have your advice in mind for the future commits.

PR #1127, opened 2021-12-06
Ultica "initial weather support"

int-ua:
#### Summary
Ultica "initial weather support"

#### Content of the change
image

#### Testing
Tested in the game but didn't make any screenshots

#### Additional information
https://github.com/CleverRaven/Cataclysm-DDA/pull/27947#issuecomment-459248944


int-ua:
At first I wanted to have common backgrounds in separate files and then just didn't remove the JSON that is redundant now

PR #1133, opened 2021-12-08
[Smap] some dirty changes to generic terrain

vetall812:
#### Summary
Smap "some dirty changes to generic terrain"

#### Content of the change

color and shape changes for:
forest
river
swamp
trail
road

#### Testing

image
image


int-ua:
The overall style of this tileset is matching pixels to map tiles but the roads in this PR are deviating from that style. I see you've tried to match topographic maps style of your country but is US they are a bit different and the game happens in US.

https://www.usgs.gov/programs/national-geospatial-program/us-topo-maps-america

example

example

curstwist:
I think the water should use two tones since knowing the depth may be important to the player.  Thick forest could be more distinctive.

int-ua:
```
road_nesw_manhole_old.png
road_old.png
```
are added as IDs by compose.py even though they are not used

vetall812:
> ```
> road_nesw_manhole_old.png
> road_old.png
> ```
> 
> ~are added as IDs by compose.py even though they are not used~ wait, at least the first one is explicitly defined, are these correct?

Not sure, i have no such warnings when composing.

vetall812:
Marked as draft, so it'll be merged just after all generated infrastructure.

Fris0uman:
Is this ready to go?

PR #1138, opened 2021-12-16
Hollow Moon docs

secretstamos:
#### Summary

HM "Add style guide"

#### Content of the change

Adds the hollow moon style guide and puts it in its own folder. Moves Utica docs into its own folder.


Fris0uman:
the `how-to` folder should remain where it is

secretstamos:
> the `how-to` folder should remain where it is

Yeah my bad, didn't read it

Fris0uman:
There's still some file moving and changes that should not have happened likme moving installation.md and summary.md

Can you just add an HollowMoon folder into the style folder and leave everything else untouched? It would be easier to review and create less change.

PR #1160, opened 2022-01-08
HM more sprites

nexusmrsep:
#### Summary
HM "more sprites"

#### Content of the change
spites names as commit names

#### Testing
image
image
image
image

#### Additional information
Special thanks to @secretstamos


Fris0uman:
A more informative title would be good. I know it's not easy especially if the sprite don't have any evident relation between them but every PR can be called "more sprites" and that doesn't make a very nice changelog

PR #1239, opened 2022-02-15
Test fire barrel for field sprites custom to furniture

bombasticSlacks:
#### Summary
Ultica "Test fire barrel for field sprites custom to furniture"

#### Content of the change
vetal's fire barrel


bombasticSlacks:
This was just till Vetal could add it themselves

PR #1252, opened 2022-02-27
msx: remove out of place white background

casswedson:
#### Summary
MSX "remove out of place white background"

#### Content of the change
Not sure how this ended up like this, it's supposed to be transparent, so here it is.

#### Testing
Screenshot from 2022-02-27 09-39-15
Screenshot from 2022-02-27 10-06-02


Fris0uman:
Thanks for the fix. This is a pretty weird mess up X)

PR #1292, opened 2022-03-25
[Ultica] Fake item sprite

vetall812:
#### Summary
Ultica "Fake item sprite"

#### Content of the change
Some items wouldn't appear in game as is. Aerosol paint is an example.
But my wish is every item have a sprite, and appear in the Cataclysm Guide with image.

#### Testing
image


Fris0uman:
can you encounter that stuff in game without debug?

Faalagorn:
> can you encounter that stuff in game without debug?

It's now merged, but I think not, though having the one for debugging (and occasional frickup) is a welcome addition, especially to make sure it shouldn't appear. :)

PR #1338, opened 2022-05-08
Fix get previous tag action in automatic releases

Fris0uman:
#### Summary
Infrastructure "Fix get previous tag action in automatic releases"


Fris0uman:
Obsoleted by #1339

PR #1342, opened 2022-05-10
fix: msx artifact name

casswedson:
#### Summary
None

#### Content of the change
in preparation for automatic tileset updates for the CDDA repo, the
release artifacts need to be named exactly as the folders in the
`gfx/` folder of CDDA

in CDDA `MshockXotto+` as a release asset here `MShockXotto+`

### DIFF
--- a/.github/workflows/make_release.yml
+++ b/.github/workflows/make_release.yml
@@ -89,7 +89,7 @@ jobs:
             artifact: Larwick_Overmap
             tileset_dir: gfx/Larwick_Overmap
             compose_args: --use-all
-          - name: MShockXotto+
+          - name: MshockXotto+
             artifact: MShockXotto+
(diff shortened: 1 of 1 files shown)


.github/workflows/make_release.yml
```diff
@@ -89,7 +89,7 @@ jobs:
             artifact: Larwick_Overmap
             tileset_dir: gfx/Larwick_Overmap
             compose_args: --use-all
-          - name: MShockXotto+
+          - name: MshockXotto+
```
Fris0uman, line 92: I only realised after merging it but it's the `artifact` entry that needs changing here, see lines 133 - 134

casswedson: Sorry about that, turns out I am blind

PR #1366, opened 2022-06-06
Update Credit.txt of MSX

chaosvolt:
#### Summary
MSX "Update credit.txt of MSX"

#### Content of the change

Credits file of MSX was missing a contributor listed. Literally my first contribution to DDA was https://github.com/CleverRaven/Cataclysm-DDA/pull/11707, which was when the Mshock set was developed into MShock Modded, which was used as the base by Xotto (source: https://discourse.cataclysmdda.org/t/20170216-big-sprite-update-mshock32xottoplus-tileset-oversprite-tileset/11301).

#### Testing

N/A

PR #1368, opened 2022-06-06
[MSX] Improve attribution of pre-migration artists

Fris0uman:
#### Summary
MSX "Improve attribution of pre-migration artists"

#### Content of the change

Improves attribution by linking to the main repo history
Obsoletes #1366


I-am-Erk:
Don't feel you have to do this immediately but we should probably look into the same format for all the legacy tilesets we've included.

Fris0uman:
I'll open an issue about it, and try to sort it out soon

PR #1377, opened 2022-06-10
msx: ft. bone and tainted bone

casswedson:
#### Summary
None

#### Content of the change
removes a couple of ultica fillers for modified versions of the
already very good looking bone shiv sprite

#### Testing
Unsure of how to indicate the rottenness of the tainted bone; just made it a little more yellow, works for now
Screenshot from 2022-06-09 19-12-30


Fris0uman:
Maybe make them greenish instead of yellowish, that way it will look more clearly zombie related I think

casswedson:
I added a little green and over all the thing is a bit darker
Screenshot from 2022-06-10 20-17-25

PR #1378, opened 2022-06-10
msx: update ultica fillers

casswedson:
#### Summary
None

#### Content of the change
Update some fillers from ultica; such that they aren't so much fillers but become more like the rest of the tileset

Mostly consisting of recolors and resizes, less colors in places a couple made more saturated; generally following other recent sprites added my other folks - made the crowbar a little bigger

#### Testing
Screenshot from 2022-06-09 23-45-24
Screenshot from 2022-06-09 23-54-41


Fris0uman:
Maybe make the needles bigger? Like instead of a small needle and a string just have a big needle? Or keep the various string colours, it makes the needles easier to distinguish

casswedson:
I'd make a case for not pushing sprites to be bigger if possible, needles are small; it makes sense for they to have a small sprite, the string complements them as they are

I wouldn't keep the string color differences, if the color of the needles themselves don't make distinct enough we should change that color instead

by the way this iteration is better than before, thank you
Screenshot from 2022-06-10 12-04-57

PR #1439, opened 2022-07-05
First try on fixing FANGS display on MUZZLE_LONG

Stadler76:
#### Summary
MSX "Fix for the white bar with FANGS on MUZZLE_LONG"

#### Content of the change

With fangs on at least the reptilian muzzle the result was a white bar on the male version and on the female version the fangs were one pixel too high.

#### Testing

None so far. First pull request for a first test.

#### Additional information
Current version (male on the left):
male lizard female lizard


Stadler76:
What I did so far is:
1. Changed the mouth in MUZZLE_LONG from alternating between white and light green-ish to a light gray bar as in shaded tooth row vs. attempting to simulate the gaps between the teeth somewhat. The old one led to that white bar on the male version, because the fangs are white and the teeth surrounding it were white, too.
2. Moved the female fangs one pixel down (Now they are the exact same, as the male version)
3. Made them 2 pixels high to 'emulate' them poking out of the mouth.
4. Moved the fangs apart by one pixel.

PR #1443, opened 2022-07-09
fix: isomsx composer

casswedson:
#### Summary
None

#### Content of the change
I noticed there were two identical msx composer workflows; one is intended to compose the iso version of msx, I fixed it.

good thing someone before me made the workflows to be pretty easy to edit

#### Testing
Download the generated artifact and load a game
https://github.com/casswedson/CDDA-Tilesets/actions/runs/2642769592
Screenshot from 2022-07-09 18-49-06


Fris0uman:
Thanks for the fix, I don't know how I managed to not do those edits when I added the workflow U_U

PR #1444, opened 2022-07-10
[NeoDays] 5 terrain sprites

Golfavel:
#### Summary
NeoDays "A few minor terrain additions"

#### Content of the change
Adds:
- Unique sprite for alien grass
- Unique sprite for pedestrian crosswalk
- "Unique" sprite for less than nothing
- Solid earth sprite taken from RetroDays

Changes:
- Open air sprite changed to nothing/completely transparent instead of light blue

#### Testing
Before:
befterr
befair

After:
aftterr
aftair
I didn't crop this one too well, but the far right would be light blue normally. I don't believe the difference between what is visibly close and far is of value.


Fris0uman:
Retrodays Could probably also use some sprite for the portal storm stuff

PR #1446, opened 2022-07-13
[RetroDays] Isherwood's forests, roads, and fields

### LINKED ISSUE #1433: [RetroDays overmap] Isherwoods specials are missing tiles
image

I am unfortunately not aware of `looks_like` existing for overmap terrain, so it would be great if the special forests, fields and rural roads could get  displayed like their normal variants. This should be relatively easy since it is just aliasing tiles since there is no reason to not re-use the existing textures of the normal variant.

Golfavel:
#### Summary
RetroDays "Isherwood's special overmap tiles"

#### Content of the change
Isherwood's "special_forest*, "special_field", and "rural_road*" now inherit normal forest, field, and dirt road sprites.
Fixes #1433 

#### Testing
Before:
befretov
After:
aftretov

#### Additional information
There are no equivalent sprites for the other tiles in this generation (farm fields, buildings, etc), so they will stay as the currently are.

### DIFF
--- a/gfx/Retrodays/pngs_tiles_10x10/overmap/dirt_road/dirt_road.json
+++ b/gfx/Retrodays/pngs_tiles_10x10/overmap/dirt_road/dirt_road.json
@@ -1,5 +1,5 @@
 {
-    "id": "dirt_road",
+    "id": [ "dirt_road", "rural_road"],
     "fg": [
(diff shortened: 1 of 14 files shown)


Golfavel:
If I remember to I'll look into it

PR #1477, opened 2022-08-08
[Infrastructure] Compose wrapper fix

vetall812:
#### Summary
Infrastructure "Changed default behavior and added support for microsoft store python installation"

#### Content of the change
1) Changed default behavior, now script will update the game directly. /d key also changed - now it allow to compose tileset into different folder
2) Added ability to launch not only python launcher (py) but also python that installed from microsoft store.


vetall812:
@Fris0uman , could you please merge it?

PR #1483, opened 2022-08-14
msx: indoor plants

casswedson:
#### Summary
None

#### Content of the change
Replace two old sprites with stripped down versions, more simple, less colors, a little bolder
Screenshot from 2022-08-14 16-16-53
Screenshot from 2022-08-14 16-40-38

#### Testing
I think this is as much as I can get away with taking something and pretty much converting it into an icon; the plants have little shadows so that should help


Fris0uman:
I think they could do being less pastel, maybe increase the saturation a bit

casswedson:
That works, these are colors from a triffid and a wasp
Screenshot from 2022-08-15 18-06-51

PR #1490, opened 2022-08-20
msx: resize aluminum cans

casswedson:
#### Summary
None

#### Testing
My aluminum cans were a little small, or I guess some of my other items are a little big, but I'd like to make these bigger

#### Additional information
Bef & After
Screenshot from 2022-08-19 19-57-53
Screenshot from 2022-08-19 20-02-55


Fris0uman:
It's better to be on the "a little too big" side for MSX, makes things easy to identify

PR #1505, opened 2022-08-27
msx: simpler cursor

casswedson:
#### Summary
None

#### Content of the change
Make the cursor simpler, a square-ish yellow line, also recolor the item highlight icon, I use this yellow in a lot of places, consistency is good

#### Testing
Screenshot from 2022-08-26 17-54-09


gfx/MShockXotto+/pngs_tiles_32x32/misc/highlight.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "highlight",
```
Fris0uman, line 2: Make one json file with an array  in the id field listing all the uses of cursor, including `cursor` right now with the json setup this way there is no sprite for cursor

PR #1507, opened 2022-08-27
HM Revisions Round 6

secretstamos:
#### Summary
HM "Revisions Round 6"

#### Content of the change

More sprite and item revisions

#### Testing

Sample image of some of the new walls/foors. Please see the changed items for more examples
image


Fris0uman:
Don't duplicate sprites, looks like NPC and player are using the same sprite so just add the json to assign the sprites to both instead of duplicating the sprites themselves

secretstamos:
> Don't duplicate sprites, looks like NPC and player are using the same sprite so just add the json to assign the sprites to both instead of duplicating the sprites themselves

done

PR #1544, opened 2022-09-06
Hm revisions round 7

secretstamos:
#### Summary
HM

#### Content of the change

Handful of new additions and revisions, including new wall icons. See below and the changes:

#### Testing

image


Fris0uman:
You should use the linter from the dda repo to format the json http://dev.narc.ro/cataclysm/format.html
Also you still have some files called `generic` in f_wreckage/Variations, it should iether have a useable name or go into a scratch folder outside the folders use for composing

PR #1557, opened 2022-09-08
Easy to use, self-contained tileset update script

mlange-42:
#### Summary

Easy to use, self-contained tileset update script

#### Content of the change

Rework `updtset.cmd` and add helper scripts to artists can set up their environment and run updates without needing to touch the command line.

Further, it is not necessary anymore to modify `updtset.cmd` for the local setup. Once environment variable `CDDA_PATH` is set, it should just work out ofthe box with double click. A helper for setting up the variable is also provided.

* [x] Use environment variables `CDDA_PATH` for the game path, and `CDDA_TILESET` for the tileset to be updated
* [x] Add two helper scripts to set these variables interactively
* [x] Modify `updtset.cmd` to let users select game path and tileset interactively if
  * not available as as environment variable (both), and
  * not given as script parameter (tileset)
* [x] ~~Move `compose.py` from the CDDA main repo into this repo, folder `tools`~~
  * ~~(required as the pre-compiled game does not contain that file)~~
* [x] Add a script to set environment variable `LIBVIPS_PATH`
* [x] Modify `compose.py` in the main game repo to include `LIBVIPS_PATH` into library search path (see https://github.com/CleverRaven/Cataclysm-DDA/pull/60923)
* [x] ~~Include `compose.py` from the main game repo into binary distributions~~
* [x] Add a copy of `compose.py` to folder `tools`
* [x] Document the new workflow for setting up (see [doc/installation_windows.md](https://github.com/I-am-Erk/CDDA-Tilesets/blob/33430ef/doc/installation_windows.md))

#### Testing

Tried it out in it's different variations.

#### Additional information

Usage without touching Command line, fixed tileset:

1. Drag&drop game directory onto `set_game_path.cmd`
2. Drag&drop `libvips` directory onto `set_vips_path.cmd`
3. Double click `set_tileset.cmd`, and choose one of the presented options
4. Double click `updtset.cmd`, should work without a hassle

Usage without touching Command line, flexible tileset:

1. Drag&drop game directory onto `set_game_path.cmd`
2. Drag&drop `libvips` directory onto `set_vips_path.cmd`
3. Double click `updtset.cmd`, and choose one of the presented tileset options

> Note: Requires inclusion of `compose.py` from the main game repo into binary distributions, in order to make it work out of the box for artists that work with precompiled binaries.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,3 @@
+# CDDA-Tilesets documentation
+
+The documentation is best viewed in it's [**rendered version**](https://i-am-erk.github.io/CDDA-Tilesets).
--- a/doc/SUMMARY.md
+++ b/doc/SUMMARY.md
@@ -1,6 +1,7 @@
 # Summary
 
 - [Installation & Building](installation.md)
+  - [Windows guide](installation_windows.md)
 - [Style guidelines](style/summary.md)
     - [General](style/general.md)
     - [Items](style/items.md)
--- a/doc/installation_windows.md
+++ b/doc/installation_windows.md
@@ -1,0 +1,53 @@
+# Windows guide
+
+Guide for Windows users, without touching the Command Line.
+
+## Install requirements
+
+**Python**
+
+Install Python 3 from [python.org](https://www.python.org/downloads/windows/).
+
+During installation, check "Add Python to PATH".
+
+**libvips**
+
+Download the latest libvips distribution from [libvips.github.io](https://libvips.github.io/libvips/install.html)
+(get vips-dev-w64-web-#.#.#.zip NOT vips-dev-w64-all-#.#.#.zip).
+
+Extract the files somewhere.
+
+**Cataclysm-DDA game**
+
+The easiest way to get the game is to use the [CDDA Game Launcher](https://github.com/remyroy/CDDA-Game-Launcher).
+With the launcher, install the latest **experimental** release.
+
+## Setting up paths
+
+For composing tilesets, some path must be known to the respective scripts.
+This section describes the most easy drag & drop approach.
+
+In `CDDA-Tilesets`, go into folder `tools`.
+
+1. Copy `set_game_path.cmd` into the game's folder, and double-click it.
+OR: Drag & drop the game folder onto `set_game_path.cmd`.
+   > Note: This sets the environmental variable `CDDA_PATH`.
+
+2. Copy `set_vips_path.cmd` into the vips folder (e.g. `C:\vips-dev-x.xx`), and double-click it.
+OR: Drag & drop the vips folder onto `set_vips_path.cmd`.
+   > Note: This sets the environmental variable `LIBVIPS_PATH`.
+
+3. Optional: To set a tileset to compose permanently, double-click `set_tileset.cmd` and select the desired tileset.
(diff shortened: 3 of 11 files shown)


Fris0uman:
The issue with moving compose.py here is that it's going to desync with the one in the main repo

mlange-42:
> The issue with moving compose.py here is that it's going to desync with the one in the main repo

Yeah, I am aware that this is not an easy decision. Moving instead of copying would avoid desync. But I start to think that the main repo *is* the right place for the script.

So maybe a solution would be to include `compose.py` in the game distribution. `updtset.cmd` could use the script from the game directory then, and it would also work for users that downloaded a build instead of cloning the repo.

What do you think?

Fris0uman:
Users could get it by going to https://raw.githubusercontent.com/CleverRaven/Cataclysm-DDA/master/tools/gfx_tools/compose.py and copy/pasting it into a new file

mlange-42:
> Users could get it by going to https://raw.githubusercontent.com/CleverRaven/Cataclysm-DDA/master/tools/gfx_tools/compose.py and copy/pasting it into a new file

In addition to making usage as easy as possible, one of the aims of this PR is to get rid of the necessity to make changes the repo(s) just to get it set up. Firstly, this is one more step to set up, secondly it clutters the repo's working copy and users have to constantly take care to not accidentially commit `compose.py` (or currently their changes to `updtset.cmd`).

Are there any reasons not to include it into the distribution?

vetall812:
As I talked to some artists unfamiliar with github I think that moving compose.py to tileset directory could be a better solution. As it will avoid some ...lets say "difficult" steps to set up and maintain infrastructure. There may be some reasons to keep it in main repository. But it is a script that work with tilesets.

mlange-42:
Have to withdraw my own argument...

> Moving instead of copying would avoid desync.

@Fris0uman probably referred to new game features that need edits to the script, which would then be in another repo/PR.

mlange-42:
Reworked to work with the script in the game directory instead of moving the script here. Requires inclusion of the script in binary distributions. Will make a PR in the game repo for it.

Fris0uman:
> users have to constantly take care to not accidentially commit compose.py (or currently their changes to updtset.cmd).
We could add to gitignore. The script is small enough that putting it in the game is not an issue but it really doesn't make sense. Trying to make it easier to contribute art is a great idea, but asking people to download one file is really not a hard bar to pass

mlange-42:
> > users have to constantly take care to not accidentially commit compose.py (or currently their changes to updtset.cmd).
> 
> We could add to gitignore. The script is small enough that putting it in the game is not an issue but it really doesn't make sense. Trying to make it easier to contribute art is a great idea, but asking people to download one file is really not a hard bar to pass

Adding it to .gitignore is a good step. However, it would require an additional manual action by the user to get it into sync after a pull or game update. 🤔

Fris0uman:
I guess we can have a copy of compose.py in this repo. It is easier for us to update it than it would be for artists than don't want to touch github

mlange-42:
> I guess we can have a copy of compose.py in this repo. It is easier for us to update it than it would be for artists than don't want to touch github

Not my favorite solution, but changed accordingly.

PR #1561, opened 2022-09-11
msx: more cursor sprites

casswedson:
#### Summary

Include a couple more cursor sprites + tweaks to the size of the previous cursor

#### Content of the change
A highlight indicator that shows up in many places like when you pick items or try to examine multiple objects, I would say necessary addition cause having a sprite for specific things is cool. a new iteration of the cursor the other one was a little small and tiled in a very much unintentional way; I did mean for it to tile but not as much an if anything this should stand out a lot
Screenshot from 2022-09-10 18-53-50
Screenshot from 2022-09-10 19-09-29

very much stolen from ultica

also I don't know what `line_target` is or where it shows up, but there is a sprite for that


Fris0uman:
I'm not convinced by the smaller cursor for highlight. What's good with the square cursor compared to the circular one we had before is that it's less obtrusive and doesn't cover stuff on the tile, a smaller curosr in the midle of the tile does though and I don't think it improves the situation.

casswedson:
that's a fair point, how about something like this an in-between approach
Screenshot from 2022-09-11 09-45-27

gfx/MShockXotto+/pngs_tiles_32x32/misc/cursors/cursors.json
```diff
@@ -0,0 +1,16 @@
+[
+  {
+    "id": [ "line_trail", "cursor" ],
+    "fg": [ "cursor" ]
+  },
+  {
+    "id": "highlight",
+    "fg": "highlight",
+    "bg": ""
+  },
```
Fris0uman, line 10:
```suggestion
```
there's no need for this since the sprite is named like the id

PR #1566, opened 2022-09-11
[Ultica ISO] Sidewalk

mlange-42:
#### Summary
Ultica ISO sidewalk

#### Content of the change
Add sprite and scratch file

#### Testing

Tried in-game

#### Additional information

t_sidewalk

grafik


Fris0uman:
you don't need the json if the fg and id are the same

mlange-42:
> you don't need the json if the fg and id are the same

Oh, really? Was not aware of that.

mlange-42:
> you don't need the json if the fg and id are the same

Done, removed the JSON file.

PR #1580, opened 2022-09-14
[Ultica ISO] small fixes of offset

vetall812:
#### Summary

fixed terrain offsets
fixed sheets offsets

#### Testing
image


gfx/UltimateCataclysmIsometric/pngs_iso_half_44x44/furniture/f_table/f_table.json
```diff
@@ -1,6 +1,7 @@
 {
   "id": "f_table",
   "fg": "f_table_unconnected",
+  "height_3d": 22,
```
mlange-42, line 4: This should be 20.

PR #1582, opened 2022-09-14
[Ultica ISO] new temporary player character

vetall812:
#### Summary
As Ultica-ISO is a tileset based on the real proportions I think that player character should be represented with temporary image. So artist can see real proportions of the world. 

#### Content of the change
removed everything related to PC
added test image

#### Testing
image
image


Fris0uman:
You're going to remake every overlay on the basis of a new character sprite?

vetall812:
> You're going to remake every overlay on the basis of a new character sprite?

Unfortunately - yes (
As this tileset is tied to real sizes - player character and all overlays has to be redone.

PR #1584, opened 2022-09-16
HM revisions round 10

secretstamos:
#### Summary
HM

#### Content of the change

Bunch of new sprite additions for HM

#### Testing
image


Fris0uman:
Seeing the json for mutagen and iv mutagen reusing the same sprite for many different item I think it would be good to go make a PR to the game repo to set the mutagen to all `looks_like` the same one, that way we don't need to make the json file in the repo *and* it might make work easier for other tilesets

secretstamos:
> Seeing the json for mutagen and iv mutagen reusing the same sprite for many different item I think it would be good to go make a PR to the game repo to set the mutagen to all `looks_like` the same one, that way we don't need to make the json file in the repo _and_ it might make work easier for other tilesets

Sure, I noticed the same thing with quite a lot of things, namely jewelry and samples too.

PR #1588, opened 2022-09-16
[MSX] State flags and 2 shot special

Coligad:
#### Summary
[MSX] State flags and 2 shot special 

#### Content of the change

- All state flags and their worn overlay
- 2 shot special and wielded overlay
- 5L gold ingot

#### Testing
image


Fris0uman:
So this is kinda all wrong: you need to have only one sprite per file, the sprite needs to be named with the correct id (which is thebid of the thing it represents) that way you don't need json and you need to give the pull request a name that describe what your doing and fill the description template
Other than that the sprites look nice

Coligad:
> So this is kinda all wrong: you need to have only one sprite per file, the sprite needs to be named with the correct id (which is thebid of the thing it represents) that way you don't need json and you need to give the pull request a title that describe what you re doing and fill the description template Other than that the sprites look nice

How will I do that?

PR #1592, opened 2022-09-17
[Ultica-ISO] Rework for 48px (Ultica character x2)

mlange-42:
#### Summary

Rework Ultica-ISO to use 48px tile width, with Ultica character and monsters scaled up by factor 2.

#### Content of the change

* Create new tileset `Ultica_iso`; leave old `UltimateCataclysmIsometric` there for a while for reference
  * CI fails due to newly added tileset
* Create basic floor and wall sprites
* Symlink Ultica character, overlays, items and monsters
* Add basic templates
* Update `compose.py` to latest version (https://github.com/CleverRaven/Cataclysm-DDA/pull/61061)

Note that this is was discussed and is agreed with the primary tileset maintainer @vetall812.

#### Testing

Tried in-game. See screenshot

#### Additional information

In-game screenshot:

grafik


mlange-42:
CI fails as expected during comparison with non-exithing tileset on `master`.

Fris0uman:
Why make yet another version of the same tileset?

mlange-42:
> Why make yet another version of the same tileset?

Besause after some testing and discussions regarding size relations, we realized that the 44x22 "version" would require redrawing of all Ultica character stuff scaled up py 1.8. With 48px width, however, Ultica character and monsters can be used simply scaled up by factor 2, which the renderer can do without artifacts.

So, this is not yet another version, it will replace the 44px version. The few things that are there will be migrated here, and the old folder will be removed.

Also, `UltimateCataclysmIsometric` (old folder) is a very inconvenient name, therefore the new folder.

Further, see the reaction by the maintainer/only other contributor. So this is agreed.

vetall812:
> Besause after some testing and discussions regarding size relations, we realized that the 44x22 "version" would require redrawing of all Ultica character stuff scaled up py 1.8. With 48px width, however, Ultica character and monsters can be used simply scaled up by factor 2, which the renderer can do without artifacts.

Not only character, but monsters too. With this new base monsters and even 32x32 items can be used.

> Further, see the reaction by the maintainer/only other contributor. So this is agreed.

Thank you for the detailed answer. I'm completely agree.

Fris0uman:
You could change the folder name of the other version, it seems unecessary to make a new folder instead ofchanging the old one. It doesn't really hurt anyting to do that, but I feel like it'd be simpler to edit the existing tileset directly instead of making a new one and then deleting the old one

mlange-42:
> You could change the folder name of the other version, it seems unecessary to make a new folder instead ofchanging the old one. It doesn't really hurt anyting to do that, but I feel like it'd be simpler to edit the existing tileset directly instead of making a new one and then deleting the old one

Removed the old 44px tileset folder. Simple rework in place won't work, as some stuff also changes to other sheets. And sheet names changed anyways.

If you squash the merge, we will see if Git manages to see anything as renamed.

PR #1597, opened 2022-09-21
[Ultica-ISO] Symlink Ultica trees

mlange-42:
#### Content of the change

Use Ultica trees in Ultica-ISO via symlinks. Replace background by ISO grass.

#### Testing

Tried in-game

#### Additional information

grafik


Fris0uman:
the trees don't get scaled up?

mlange-42:
> the trees don't get scaled up?

Exactly. This is intentional, as they would just obscure so much. For iso and realistic size relations, trees should probably just the trunk, and maybe semi-transparent canopy. But there is IMO also the need for a more general solution to the occlusion problem in iso.

So for now, we just have tiny but recognizable trees.

PR #1612, opened 2022-09-25
msx: some overlays as items

casswedson:
#### Summary

modify existing character clothing overlays to be used as items
#### Testing
mostly hats
. removes ultica fillers and some legacy sprites
image
I wouldn't want to add full back outlines to some of these cuz I feel it ruins them


Fris0uman:
From the screen shot some hats are missing parts of the black outline

PR #1613, opened 2022-09-25
[Ultica-ISO] concrete floor

vetall812:
#### Testing
image


Fris0uman:
Isn't 32 variant a bit overkill? X) Also youdon't need `bg` in the json if there's no bg

vetall812:
Just a bit. there would be grass, snow. Anything not higher than 11px.
yeah, `bg` come from template, will remove it.

PR #1618, opened 2022-09-26
Add Python script to slice into numbered variants

mlange-42:
#### Summary
Add Python script to slice into numbered variants

#### Content of the change

A new script `slice_variants.py`, working like `slice_multitile.py`, but creating numbered sprites and working for multitiles of arbitrary size.

Intended to make it easier to make sprite variants.

#### Testing

Tried on 4x4 and 5x5 iso templates.

#### Additional information

Could have added it to `slice_multitile.py`, but did not want to add too much only slightly related functionality to the script.

### DIFF
--- a/doc/how-to/autotiles.md
+++ b/doc/how-to/autotiles.md
@@ -184,3 +184,17 @@ when you want to adjust all sprites as one image:
 cd mud_tiles
 $ tools/unslice_multitile.py mud
 ```
+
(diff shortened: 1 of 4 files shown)


Fris0uman:
Please add the doc to use the script too

mlange-42:
> Please add the doc to use the script too

Done.

PR #1631, opened 2022-09-28
Another batch of item sprites (take 4)

Coligad:
#### Summary
Another tileset contribution
#### Content of the change
Added American-patterned armors

#### Test
image

image


Fris0uman:
Looking much better, it would also be nice to properly feel the template describing the PR though

Coligad:
> Looking much better, it would also be nice to properly feel the template describing the PR though

Is what I do now good?

Fris0uman:
You need to test your changes to see if they actually appear in game, also try to follow the style guide https://github.com/I-am-Erk/CDDA-Tilesets/tree/master/doc/style/MShockXotto%2B avoiding noisy sprites, using pure black outlines, trying to not make giant 32x32 sprites for things that should be small etc, especially in the case of the two hard hat here that looked like edited version of old sprites from MShock or undeadpeople

PR #1658, opened 2022-10-11
[MSX] Layering for hoistable flags

mlange-42:
#### Summary
Layering for hoistable flags in MSX

#### Content of the change

Add layering for flags to implement https://github.com/CleverRaven/Cataclysm-DDA/pull/61551

#### Testing

Tried in-game (see screenshot).

#### Additional information

grafik


Fris0uman:
does it work the same with the state flag and pride flags variants?

mlange-42:
> does it work the same with the state flag and pride flags variants?

Jep! That's the nice thing that they don't need to be listed separately. (Had to implement this, did not work with variants at all before).

grafik

PR #1707, opened 2022-11-20
feat: mainlined whaley's hair mod

sonphantrung:
#### Summary
Mainline Whaley's Hair and Tattoo mod

#### Content of the change
Add relative tilesets for the [PR](https://github.com/CleverRaven/Cataclysm-DDA/pull/62297)

#### Testing
Works, see the PR above


Fris0uman:
So same as the other PR without a license or at least the original author agreeing to release those sprite under cc by sa we can't use it. Also you add the original author as co author here

Fris0uman:
looks like the fix to the licence is not happening

sonphantrung:
Oh, today I was going to amend the commit!

PR #1821, opened 2023-01-27
[MSX] Aluminum stepladder

gettingusedto:
#### Summary
Sprite for aluminum stepladder

#### Content of the change

Sprites for aluminum stepladder

#### Testing
The animation appears to be right
Screenshot_20230128-035548

#### Additional information
I hope I did the animation right.


Fris0uman:
the item version should probably have black instead of grey pixel around the holes, they are outline pixels so they should be full black

PR #1867, opened 2023-02-23
[Msx]-flowers!

Dandy-boy:
#### Summary
[MSX] - lots of plants

#### Content of the change
24 sprites encompassing most of the furniture-flora group, a bit too many for me to want to name here.

#### Testing
python isn't real and cannot hurt me


Fris0uman:
Can you put the flowers in a `flower` folder? This was cleaner I think

Dandy-boy:
> Can you put the flowers in a `flower` folder? This was cleaner I think

I opted not to because not all of them are flowers, a plants-furniture folder would work great but since i use the web interface my only solutions to moving them now would be to delete and re-submit them since it was designed by literal goofy grey aliens

PR #1890, opened 2023-03-07
[ASCII_Overmap] Delete unnecessary files

pedestrian4:
Deleted all of unnecessary parts that already covered by fallback.png file.

#### Summary

Deleted all of unnecessary parts that overlap with the fallback.png file.

#### Content of the change
Deleted all of unnecessary parts that overlap with the fallback.png file. also fully added tiles from 'data/mods'

#### Testing
image


Fris0uman:
You need to also delete the json that reference the sprite your removed. See the error in the failed test for the list of files

PR #1896, opened 2023-03-09
[NeoDays] Better effect overlays

Wreckstation:
#### Summary
I feel like the symbols representing each effect right now are just letters right now. (running, bleeding, etc) could be done more creatively, so I designed small simple symbols that could represent these effects. I also wanted to make the hostile_sees_player look distinct from the other sees_player, even in gray-scale.

#### Content of the change
Edits the sprites for hostile sees player, crouching, running, prone, and bleeding

#### Testing
image

#### Additional information
Additional testing against different backgrounds and also multiple effects at once could be done to further test visibility.  I only used the currently existing colors of the sprites, so they can always be improved.


Fris0uman:
I'm not sure this is clearer to understand, letters are pretty straitghforward

Wreckstation:
I can accept this, however I think that bleeding and hostile sees_player still could use some tweaks. Would a new pull request with only those tile changes be acceptable?

Fris0uman:
You don't need to make a new PR, you can edit this one. So leave the movement mods as they were and just keep your change to bleeding and the two sees player

PR #1950, opened 2023-04-11
[Neodays] Better contrast for brown hair

Vaskritaya:
#### Summary

Changes color of brown hairstyles for better contrast
#### Content of the change

Brown hair is currently incredibly hard to distinguish from one of the current skintones (light brown), and is just generally a bit of a strange shade for brown hair. I've tried to adjust the color of brown hair to have better contrast with all skintones for readability. 
#### Testing

before:
image
after:
image

#### Additional information
none


I-am-Erk:
I see what you're going for but I think this is mostly a lateral change... The new one doesn't contrast as much with the background

Vaskritaya:
> I see what you're going for but I think this is mostly a lateral change... The new one doesn't contrast as much with the background

yeah, I was kinda thinking the same thing. I'm gonna mess around with it more.

Vaskritaya:
Changed it to a much more saturated color to work better with the space limitations. Has better contrast with the background and darkest skintone than my previous version, and still manages to be distinguishable from all skintones and looks more like brown hair in my opinion. Thoughts @I-am-Erk ?

I-am-Erk:
I think it's fine, sorry, just haven't been around much.

PR #1960, opened 2023-04-20
[NeoDays] 3d walls (WIP)

Wreckstation:
#### Summary
You can see the sides of walls. 

#### Content of the change
Changed wall tiles to use proper 4x4 multitile:
* log/wood wall
* white/purple/blue/red/green/yellow walls
* concrete wall
* rock wall
* brick wall (now visually distinct from a wooden wall)
* chain link fence
* barbed wire fence

#### Testing
image

#### Additional information
I am still looking for any walls that I have missed. I am also in the process of figuring out how to 3d-ify open secret doors, but not high priority right now bebcause i dont think i've ever seen one in actual play


LovamKicsiGazsii:
I'm not a fan of this personally. Neodays/Retrodays portrays almost everything in a 2d style.  This deviates from the general design of the tileset.

Vaskritaya:
Yeah, it does look quite neat, but I'm not sure it works with the overall vibes of neodays without touching basically all terrain and a lot of furniture.

Wreckstation:
I'm not sure if neodays and retrodays need to look so similar if the other still exists and is still playable. but also fair. if this isn't getting merged I'll take the hint and start making my own tileset haha

I-am-Erk:
It might be worth forking the set for a change this major. We can then revisit if we need so many similar sets

Wreckstation:
otay

PR #2008, opened 2023-06-02
Tileset changes for 3d vision

Rewryte:
#### Summary
Tileset changes for https://github.com/CleverRaven/Cataclysm-DDA/pull/65738

#### Content of the change
This should only be merged if https://github.com/CleverRaven/Cataclysm-DDA/pull/65738 is.

The PR introduces 3D vision.  Some tilesets require minor changes as follows:

- `t_open_air` and variants should either be unassigned or semi-transparent.  A blank sprite will make z-levels indistinguishable from each other.  I opted to add simple sprites because any fallback eats a bit of performance.  Simply change them when you have something better.
- `tr_ledge` needs to be assigned a blank sprite to prevent a fallback sprite from being drawn.  I reuse the old `t_open_air` for this.

Affected tilesets are BrownLikeBears, Chibi_Ultica, MShockXotto+, NeoDays, Retrodays, and UltimateCataclysm.
My sprites are made in Krita 5.1 and saved using default settings.

#### Testing
Compose script ran successfully and modified tiles interact properly with the 3D vision PR.  Did not notice any unintended changes in comparison to the current master.

#### Additional information
Not too familiar with the workflow of this repo.  Let me know if anything blows up and feel free to edit my changes.


Fris0uman:
We try to not duplicate sprites, so if empty.png and t_open_air.png are both an empty canvas of the same size they're should be just one file

Rewryte:
> We try to not duplicate sprites, so if empty.png and t_open_air.png are both an empty canvas of the same size they're should be just one file

Thanks, I am aware of that.  I am changing t_open_air to a semi-transparent image so any sprite that should be blank can use the new emptyX.png instead.  I don't believe that similar sprites exist.  Let me know if any slipped my notice though.

Fris0uman:
I missed that open air what not empty, my bad

PR #2032, opened 2023-06-20
[NeoDays] Medical zombie variants

ravveni:
#### Summary
Adds medical zombie variants to differentiate from "regular" zombies in a crowd.

#### Content of the change
Adds:
- Sawbones ->  mon_skeleton_zombie
- Zombie medic  -> mon_zombie_medical
- Doctor Burns -> mon_zombie_medical_acidic
- Medical horror -> mon_zombie_medical_brute
- Miracle lurker -> mon_zombie_medical_regenerating


Fris0uman:
You don't need the json files when id and fg are the same. If you name the png file with the id of the thing they represent you don't need json. So just delete the json files

PR #2036, opened 2023-06-22
Isometric tileset changes for 3d vision

Rewryte:
#### Summary
Tileset changes for https://github.com/CleverRaven/Cataclysm-DDA/pull/66383

#### Content of the change
Basically the isometric counterpart of #2008.
This should only be merged if https://github.com/CleverRaven/Cataclysm-DDA/pull/66383 is.

Isometric tilesets require minor changes as follows:

- `t_open_air` and variants should either be unassigned or semi-transparent.  I opted to unassign them and let the fallback fog do the job.  Feel free to replace it if there is something better.
- `tr_ledge` needs to be assigned a blank sprite to prevent a fallback sprite from being drawn.

Affected tilesets are SmashButton_iso and Ultica_iso.
My sprites are made in Krita 5.1 and saved using default settings.

#### Testing
Compose script ran successfully and modified tiles interact properly with the 3D vision PR.  Did not notice any unintended changes in comparison to the current master.


Rewryte:
> hitbutton is missing a t_open_air sprite here, no?

Well, you're not wrong.  The 3D vision PR includes changes to the fallback system when sprites are not found.  A missing `t_open_air` will cause fog to be drawn in its place.

Rewryte:
Yeah, I guess I could.  It wouldn't be exactly the same, but after trying it out it did not turn out as bad as I thought it would.

Chiquita Waters _2023-06-24T00-27-02

PR #2063, opened 2023-07-14
Sprites for shadows, running, and bashing

Rewryte:
#### Summary
Assigns sprites for z-level shadows and running / bashing animations

#### Content of the change
Z-level shadows were added in https://github.com/CleverRaven/Cataclysm-DDA/pull/66730
A few tilesets already have their own shadow sprites so I reused them.  Made new sprites for the rest.  My sprites are made with Krita 5.1 and saved with default settings.

Running and smashing animations were added in https://github.com/CleverRaven/Cataclysm-DDA/pull/65622
I reused the `animation_hit` sprite as `bash_effective` and `fd_smoke` as both `bash_complete` and run_*.  If a tileset lacks these sprites then nothing will be added.

#### Testing
Tilesets compiled successfully.  Tested shadows, running and bashing for each modified tileset.

Shadows (UltiCa)

Bashing (MSX+)

Running (BrownLikeBears)


Fris0uman:
I'm afraid using smoke for running is going to be confusing. Right now smoke is something you need to pay attention to and avoid, it's a bad idea to reuse a sprite of a danger for something cosmetic

Rewryte:
Alright, I'll glance through the sprite sheets to see if anything else catches my eye but I'll probably just remove the run animation if I can't find anything.  If you have any suggestions, feel free to share them.

We cool with `fd_smoke` as `bash_complete`?

Fris0uman:
> We cool with fd_smoke as bash_complete?

No, it's the same issue, using the srpite for something dangerous for something that is not

Rewryte:
> No, it's the same issue, using the srpite for something dangerous for something that is not

Ok.  Got rid of that as well.

Also changed `bash_effective` to use glass crack sprites if exists.

PR #2065, opened 2023-07-18
[ASCII Overmap] Updates to the latest

pedestrian4:
#### Summary
Updates to the latest version

#### Content of the change

More than 100 overmap tiles added since 0.G


Fris0uman:
What convention are you using to name the glyphes?

pedestrian4:
I used the character's color and extended ascii code number (like black179). the older ones were already ramdomly chosen I want to change it though.

pedestrian4:
> What convention are you using to name the glyphes?

I used the character's color and extended ascii code number (like black179). the older ones were already ramdomly chosen I want to change it though.

Fris0uman:
you can if you want to, but I don't think it's necessary to change the already existign names

PR #2073, opened 2023-07-23
msx: replace ultica table

casswedson:
#### Summary
replace the table furniture that comes from ultica with something more in line with msx's style

#### Content of the change
Screenshot from 2023-07-23 12-33-37

#### Additional information
god please no more multitile furniture I am in PAIN


Fris0uman:
This does look good, but I think we still want tables to be multitile

casswedson:
well yea multitiles are very nice but I am certainly not making a multitile for this now, not anytime soon; I welcome an effort to complete it
I personally think this is a good enough replacement as is, just the one tile

Fris0uman:
just replace the disconnected sprite then

PR #2087, opened 2023-08-01
[Ultica] Ultica gun id fix

gettingusedto:
#### Summary
Fixing the ar15 and m4 carbine id for the ultica tileset

#### Content of the change

Did the same way i fix the ones at msx.

#### Testing

Screenshot_20230801-201731


Fris0uman:
why not rename the files instead?

gettingusedto:
not sure how i would do that without deleting and readding the sprite with a different name on my phone. also the json way is less of a hassle to me.

gettingusedto:
superceded by #2088

PR #2103, opened 2023-08-13
[Larwick] Backrooms

Golfavel:
#### Summary
Adds backrooms overmap sprites

#### Content of the change
Adds: backrooms floor

#### Testing
_Before_
lar_vanbef
_After_
lar_vanaft


Fris0uman:
Isn't this pretty unconfortable to look at? I'm not sure making the hwole screen bright yellow is a good idea

Golfavel:
The overmap has no use with the backrooms mod and the predominant color is yellow (albeit not as vibrant). I can change it if it's considered necessary.

Fris0uman:
Yes change it for something less agressive

Golfavel:
Done.

PR #2104, opened 2023-08-13
[Ultica] casings

vetall812:
like https://github.com/gettingusedto do for msx

#### Summary
Added casings for every caliber using BG sprite for ammo

#### Content of the change

- [ ] folders rearrangment

- [ ] casings


vetall812:
I would miss 5x50 shotguns :`(

PR #2117, opened 2023-08-23
Add Nix support

barsoosayque:
#### Summary
Add support for [nix](https://nixos.org/) in the repository in form of a flake.

#### Content of the change
Add `flake.nix` with ability to build tilesets and enter a devshell.

#### Testing
```sh
nix build .#UltimateCataclysm
cat result/tileset.txt
````
Produces a working tileset.

#### Additional information
It's what I use to comfortably work with tilesets, figured it wouldn't hurt to have this in the repo if there are other madlads who use nix. It's also possible to run reproducible nix builds in the CI now.

Ah. I should probably also add a book building package. Someday..

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -27,10 +27,9 @@ You will need:
 - Python 3
 - [Libvips](https://libvips.github.io/libvips/install.html)
 - pyvips (install it via python pip: `pip install pyvips`) 
-- [compose.py](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/tools/gfx_tools/compose.py) script from the main [Cataclysm-DDA](https://github.com/CleverRaven/Cataclysm-DDA) repository
 
 Once you have everything ready, you can build the tileset:
 ```sh
 # Assuming that you are in the root of the tileset repository
-$ python3 <path-to-compose-py-script>/compose.py gfx/UltimateCataclysm
+$ python3 tools/compose.py gfx/UltimateCataclysm
 ```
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: doc/installation.md [hunk 1]
```diff
@@ -27,10 +27,9 @@ You will need:
 - Python 3
 - [Libvips](https://libvips.github.io/libvips/install.html)
 - pyvips (install it via python pip: `pip install pyvips`) 
-- [compose.py](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/tools/gfx_tools/compose.py) script from the main [Cataclysm-DDA](https://github.com/CleverRaven/Cataclysm-DDA) repository
```
Fris0uman, line 30: why remove this line?

doc/installation.md
```diff
@@ -27,10 +27,9 @@ You will need:
...
 
 Once you have everything ready, you can build the tileset:
 ```sh
 # Assuming that you are in the root of the tileset repository
-$ python3 <path-to-compose-py-script>/compose.py gfx/UltimateCataclysm
+$ python3 tools/compose.py gfx/UltimateCataclysm
```
Fris0uman, line 34: we probably should not have a copy of compose in here since the one that actually gets worked on and updatd is in the dda repo

doc/installation.md
barsoosayque: ah well I thought we adopted the `compose.py` script. apparently not !

doc/installation.md
barsoosayque: hmm I wonder why it was added to the repo in the first place. I remember that you have to have a main cdda repo for this script

doc/installation.md
Fris0uman: So it is up to date, and last update to the one in the dda repo was also immediatly duplicated here, so I guess it's fine but it feels fragile

PR #2138, opened 2023-09-15
make use-all the default behavior when running compose.py

smileynet:
#### Summary

make use-all the default behavior when running compose.py

#### Content of the change

As use-all is the preferred outcome when running compose.py, invert the logic to instead present a `--no-use-all` argument instead, in the event that the user does not want the behavior.

#### Testing

1/ Run script with default arguments, no warning generated and all tiles included.
2/ Run script with `--no-use-all`, get `WARNING handle_unreferenced_sprites` messages

### DIFF
--- a/tools/compose.py
+++ b/tools/compose.py
@@ -6,7 +6,7 @@ a tile_config.json and tilesheet .png file(s) ready for use in CDDA.
 Examples:
 
     %(prog)s ../CDDA-Tilesets/gfx/Retrodays/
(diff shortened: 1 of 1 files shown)


casswedson:
you'll probably want to make edits to CleverRaven's compose.py the one hosted here isn't even used, also make sure to update the ci suite

Fris0uman:
Not the most usefull edit, but it doesn't seem to break anything, especially with the --use-all argument legacy

PR #2192, opened 2023-11-10
[Ultica] Add the Santa items (worn only)

pizzasgood:
#### Summary

I drew overlay sprites for the Santa clothes.

#### Content of the change

Santa clothes:
* santa_jacket
* santa_jacket_short
* santa_pants
* santa_shorts
* santa_dress_long
* santa_dress
* santa_dress_short
* santa_hat
* santa_beard
* santa_gloves
* santa_belt

#### Testing

Santa_Lineup


Fris0uman:
You don't need the json file just directly name hte spirtes with the id, like `overlay_female_worn_santa_beard` instead of `santa_beard_f`

pizzasgood:
I assumed github would send a notification when I pushed, but in hindsight I don't actually know if it works that way.  So just in case, here is a possibly redundant message to confirm that I've removed the json, renamed the sprites, and verified that everything still loads and displays properly.

PR #2197, opened 2023-11-17
[Tools] Added script that change default Git behavior on windows systems

vetall812:
#### Summary

When starting using repository on windows system it is usual that symlinks cannot be created or followed.
This script gives a link to the Microsoft site that explain how to enable symlinks and execute commands that enable symlinks in repository if its already on the system.

#### Content of the change
git_symlinks.cmd script in /tools


Fris0uman:
You probably should document this somewhere

vetall812:
Yeah, I am working on docs pages right now. Will include description there.

PR #2199, opened 2023-11-17
[Tools] Added ability to install specific versions of Python and LibVIPS

vetall812:
#### Summary

the main wrapper 'updtset.cmd' changed to install specific version of Python and lib-vips if they werent in the system.

#### Content of the change
First time tileset repository cloned allow you to run updtset.cmd and it will check for Python, pyvips module and libvips.
If any of this part absent in system script will try to download and install them

#### Testing
image

Known bugs:
on windows 10/11 user may have to manually disable app execution alias in system setting. This happens if python was installed via script.

### DIFF
--- a/tools/updtset.cmd
+++ b/tools/updtset.cmd
@@ -179,6 +179,9 @@ if errorlevel 1 (
         if errorlevel 1 (
             echo ERROR: No Python found!
             echo If you are sure that Python is installed - please check 'path' environment variable.
+            echo Script will try to install Python 3.12, try to run this script again after this.
+            echo.
+            winget install Python.Python.3.12 --disable-interactivity
             goto stop
         ) else (
             set APP=python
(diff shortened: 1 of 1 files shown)


vetall812:
About installing software:
1) Python checked by three different executables, if any of them is found script will try to use this python and will not install anything new.
1.1) If there weren't any python installed script will install Python 3.12 from 'winget' repository. This action does not require elevated rights and this version can be upgraded as any other usual installed software.
1.2) There is known 'bug' that windows doesn't respect Python installed this way and user have to disable sofware execution alias in windows settings.
2) LibVips checked by its executable, but as it is just a library it cant be upgraded/uninstalled automatically.
2.1) To solve this script will download specific LibVips archive from github and extract it into users home folder.
2.2) Script will not add path to LibVips to the permanent users PATH environment variable, instead script will create specific environment variable that would be added in the composing process. So this script either will use preexisting LibVips or extract one that will be used exclusively in this script.
2.3) If later user decide to add another instance of LibVips script will continue to use previous version. 
3) pyvips module will be added automatically to any python version except the one from the Microsoft Store. This Python version can't be used.

PR #2232, opened 2023-12-05
[MSX] Adobe walls and XE inventor blades

gettingusedto:
#### Summary
Adobe walls and 2 XE inventor blades.

#### Content of the change

Basegame
- Adobe brick wall

Xedra Evolved
- Binary Sword (item and wielded)
- megachette (item and wielded)


gettingusedto:
Done.
Screenshot_2023-12-07_10-08-54_1
looks like i did it right on the first try. it's rather surprising.

PR #2249, opened 2023-12-18
[Tools] slicers wrapper

vetall812:
#### Summary
This script allow artists to speedup slicing multitiles process.
It detects tile name, size of resulting tile, and various source files:
  - simple multitile
  - transparent multitile
  - variants for single tile
  - seasonal variants (as well as seasonal transparent variant)

Just drag and drop source folder over script shortcut.

#### Content of the change
1) to slice multitiles python script `slice_multitile.py` requires NumPy module. This module can be used with Python >= 3.7 and < 3.11 So I change `updtset.cmd` so it will install Python 3.10 instead of 3.12 (latest at this moment).
2) `updtset.cmd` will also install `NumPy` module at first run.
3) to keep repository clean from multitiles there is a new line in `.gitignore`
4) added slicemt.py script. 

#### Testing
1) Create a shortcut to the script on your desktop (for example)
Screenshot 2023-12-19 181936

2) Put the source files into the folder and drop it in repository
Screenshot 2023-12-19 182134

3) Drag and drop 'source' folder over the shortcut
Screenshot 2023-12-19 182228
Script informs you about files taken into account as well as resulting names and sprite numbers. Script will ignore existing json files, so they can be changed without fear.

4) After displaying results script will run 'updtset.cmd' so just press 'Enter' and wait a bit
Screenshot 2023-12-19 182256

Now you can update tileset in game and see how it works.

#### Additional information
You may notice shortcut to 'updtset.cmd' on my desktop - it is here just for simple sprites that do not need to slice something. Or if I change json files without touching images.


vetall812:
It would be great if someone rewrite it on Python then we can just drag-and-drop `source` folders over python script.
Bonus if this python script detect tileset und update it.
PowerShell is simple but so secure that you can't use drag-and-drop over ps1 script.

gkarfakis19:
Looks good!

Fris0uman:
Can you add documentation for this to our doc?

vetall812:
> Can you add documentation for this to our doc?

Done in #2255

PR #2255, opened 2023-12-22
[Docs] update

vetall812:
#### Summary
Moved `Ultica` style guide into separate section
Added `tools` section
Fixed links style in various documents

#### Testing

[check it here](https://github.com/vetall812/CDDA-Tilesets/tree/docs_common_update/doc)

#### Additional information
 
Should work with the rendered version too, but can't check before PR gets merged.


Fris0uman:
The skin images in ultica style guide don't work, the doc should probably point to the sprites directly jnstead of using symlinks

vetall812:
> The skin images in ultica style guide don't work, the doc should probably point to the sprites directly jnstead of using symlinks

That's because pull requests have a separate link pattern. When merged symlinks start to work.
`https://github.com/vetall812/CDDA-Tilesets/blob/docs_common_update/doc/style/UltimateCataclysm/image/skin-male.png`

I can create a separate image file for skins. But previous docs version uses a symlins, so do I. I think we should keep symlinks for now. They are the good indicator if something changed in repository.

vetall812:
> The skin images in ultica style guide don't work, the doc should probably point to the sprites directly jnstead of using symlinks

On the other hand they didn't work for old docs either. So probably separate images would be better. Let me sumbit them.

image

PR #2357, opened 2024-04-20
[Ultica] 2 ferals + zombie medic + variants by Majstek

hoasd1:
#### Summary

Adds 2 ferals, zombie medic and 2 variants by Majstek
#### Content of the change

Monsters :

- feral biker
- feral swimmer
- zombie medic

Added one variant for zombie medic but with hanging mask
Added one variant for zombie scientist but with different clothes, face and crippled leg

Changed JSON "weight" of the previous zombie scientist variants to make the crippled leg variant less common

#### Testing
2_ferals_+_medic_zombie_+_variant_by_majstek

#### Additional information
There were sprites with mask covering their mouth but I decided to not add them because both zombies have the "bite" attack


hoasd1:
First time resolving a conflict through Github website interface, The change in "mon_feral.json" is unintended ( It looks like I put both "mon_feral_swimmer_kickboard" and "mon_feral_prepper_mace" in the same bracket ) >.>
Will fix it in another submissions PR because it's more confusing after the commit :/

hoasd1:
Thinking about the unintended change, I see it's better to add the changes to a new PR.

PR #2410, opened 2024-06-12
Fix outlines + variants

Mawranth:
#### Summary
Uploaded purple and white variants. Variants for cosplay. Fixed outline on buns.


Fris0uman:
A few issues:
- The outlines are not RGB: 0 0 0
- You're pushing this pull request from your own master branch which as a rule you should never do
- Looks like at least some of the sprites are saved in indexed mode instead of RGB, we have no rule against that but it makes it a bit more difficult to fix the outline color

Fris0uman:
oups didn't mean to close it, that's the issue with working on a masgter branch it does weird stuff

PR #2489, opened 2024-08-21
[P&P] Documentation update

vetall812:
#### Summary

Added tileset documentation.


Fris0uman:
The doc should probably just be in english, we have no way to manage this project in any other language, also what's the license on the two old_map.png images?

vetall812:
> what's the license on the two old_map.png images?

old_map1.png is a part of CC0 image from https://en.wikipedia.org/wiki/Glamorgan#/media/File:Glamorgana_Atlas.jpg
old_map2.png is a part of a antique map photo, the map itself from 1900, so I think there should be no problem with it. Source: https://www.puddletownbookshop.co.uk/product/47237/An-Accurate-Map-of-Glamorgan-Shire-Drawn-from-an-actual-survey-with-various-Improvements-Illustrated-with-Historical-Extracts-relative-to-its-Natural-Products-Air-Soil-Trade-Manufacture-ampc

vetall812:
> The doc should probably just be in english, we have no way to manage this project in any other language.

I think that additional language is an option. But I can remove files of course.

Fris0uman:
Maybe just add the sources for the picture in a txt file in there for bookeeping. We can actually support non english contributors so I feel it'd a bit of false advertising to have the doc in another language, so yeah just remove it

PR #2589, opened 2024-11-30
Flag factory, national flags, layering updates

ShnitzelX2:
#### Summary

- ultica_build_flags.py will turn flat flag images into layered sprites + item sprites
- top 30 national flags by population
- updates to layering.json to support better layering

This PR should **NOT** be merged until (https://github.com/CleverRaven/Cataclysm-DDA/pull/78199) is merged.

#### Content of the change

I found Erk's flag issue (https://github.com/CleverRaven/Cataclysm-DDA/issues/60405) when I first started playing CDDA, back in March/April. I made it work on the C++ side, but putting every flag sprite into the game seemed insurmountable; each flag has effectively three required sprites.

So I made `ultica_build_flags.py`. It turns a flat flag sprite into a mildly textured image using GIMP. It provides a template for future sprites to be built in large batches using code. Building sprites in this way is very versatile -- it can do most of what GIMP can do -- and not _too_ difficult to work with, but you do need basic proficiency with GIMP.

Using this, I made the top 30 national flags by population as sort of a test case and included the outputs (along with the base textures) in this PR.

#### Testing

From top to bottom, the "_postup" national flag sprite, the dropped flag sprite, and the "_hoisted" flag sprite.

Note: I forgot to put Turkey's flag up, and the Colombian flag is misspelled in the main game so it doesn't display.
flags

#### Additional information

You can use `ultica_build_flags.cmd` to run `ultica_build_flags.py` for every desired flag in "scratch\UltimateCataclysm\items\flags". The flags are saved to an "output" folder in the same directory.

I ran `ultica_build_flags.py` on Windows, untested on Linux. It uses GIMP's built-in Python IIRC.


Fris0uman:
Adding some documentation for the new script would be nice

PR #2601, opened 2024-12-19
[DinoMod] MSX+ apatosaurus

LyleSY:
#### Summary
New tile for apatosaurus in DinoMod

#### Content of the change

PNG plus credits file with the link of the file this was modeled from, here https://upload.wikimedia.org/wikipedia/commons/thumb/4/44/Apatosaurus_louisae_by_durbed.jpg/320px-Apatosaurus_louisae_by_durbed.jpg (shortened the tail because it didn't read visually when shrunk down so much)

#### Testing

I don't have python set up on my machine right now but it should work

#### Additional information

Most important missing piece for DinoMod


Fris0uman:
There are way too many colors in that sprite, I'd prefer if we kept moving away from crunchy scaled down pngs.

LyleSY:
Ha, yeah I just saw the "three to four colors per material" guidance. This was a little higher than that. I've trimmed it down to three color options for the main body, a blue dot for the eye, and the black outline.

Fris0uman:
It needs a shadow

PR #2609, opened 2025-01-12
[Ultica] [MSX] Coffee pot by UMP40/akira.sato + remove obsolete "Cave" from Wynnter

hoasd1:
#### Summary

Adds Coffee pot by UMP40/akira.sato and remove "Cave" from Wynnter from MSX
#### Content of the change
Items :

- coffee pot

Removed "Cave.png" overmap by Wynnter because there's "cave.png" overmap recently added by GetUsedTo 
#### Testing
Coffee_pot_By_akira_sato


Fris0uman:
Why Ccave and not just cave?

hoasd1:
> Why Ccave and not just cave?

Because I wanted to do both changes in one commit
Can't remove the other identical filename ( according to default Windows explorer file setting ) on the same commit

hoasd1:
Thought I could let it go like that, but it seems it won't :l

PR #2616, opened 2025-02-11
[MSX] Pruning unused sprites

Karol1223:
#### Summary

Removed sprites for monsters that have been removed entirely (as in, not being in an in-repo mod either)
Removed the unused duplicate sprite for tadpoles


Fris0uman:
Can you edit the PR `content of changes` to list the monster removed for book keeping please ? That way it will be a bit easier to come find the sprite again if we ever feel like reusing them for something else

PR #2689, opened 2025-07-02
[Ultica] Adds some terrain art for Labyrinthine Structures

I-am-Erk:
Mostly it's just swirly blackness

#### Summary
[Ultica] Adds some terrain art for Labyrinthine Structures

#### Content of the change
Swirling chasm: Complete (multitile)
Walls and floors pending.

#### Testing
image
image
Worked hard to keep the tiling from being too repetitive.

image
I am not super happy with the catwalk tile, I think I'll draw something from scratch. This is better than the ultra repetitive diamond plate until then, though.

#### Additional information
I hope I remembered how to do this, it's been a minute


I-am-Erk:
Oh right this is not mainline dda, I don't need to wait for a merger.

PR #2708, opened 2025-07-14
Move fallbacks from main DDA repo to tileset repo

harakka:
#### Summary
Move fallbacks from main DDA repo to tileset repo

#### Content of the change
We're in process of starting to compose tilesets in main DDA repo for releases (https://github.com/CleverRaven/Cataclysm-DDA/pull/81758), as a result the copies of tilesets living in main repo are going away. There were a bunch of fallback.png's on the main repo side that  didn't exist on tileset repo side, this PR moves those over.

Even if we don't go through with the change on DDA repo, this data belongs on the tileset repo as they're part of the tilesets.


Fris0uman:
Why only those 4?

PR #2709, opened 2025-07-14
Rename debug_mon to pseudo_debug_mon in ultica

GuardianDll:
#### Summary
For a while the texture was broken after the monster was renamed
#### Content of the change
Just renaming
#### Testing
None
#### Additional information
Super unfamiliar with how tileset works, if anything else need to be done please comment


Fris0uman:
Can you do it for the other tilesets please?

PR #2813, opened 2026-01-30
[Altica] Adds Body-Length Hair, Resprites Long Over Eye and Extremely Long Hairs, Adds Eyebrows/Eyelashes to All Hair Still Missing Eyebrows/Eyelashes

TheIssueFairy:
#### Summary
- Adds Body-Length Hair
- Resprites Hairs: Long Over Eye, Extremely Long (replaces my old sprites with better ones)
- Adds gendered versions to all of my hair sprites, the only difference is 1 pixel in the eyebrows/eyelashes but this is necessary to match the other Altica hair sprites. Not having them just caused the versions of the eyebrows/eyelashes on the skin sprites to be visible, which didn't look as good as having them match the hair. The reason my sprites didn't originally have them was because I misinterpreted the sprite I was using as a reference and thought they were part of the hair.
#### Content of the change
A large quantity of hair sprites.

#### Testing
Eyelash Fix:
image
image
image
image

Hair Resprites:
image
image
image
image

New Hair (Body Length):
image
image


TheIssueFairy:
Removed the unused ungendered versions of the sprites I just changed because I thought they were causing an issue, turns out it was entirely my fault as I managed to miss part of the tileset artifact while copying it into my game. I'm keeping up my pattern of inventing a new mistake every time I make a PR.

Leaving them removed 'cause the game will use the gendered versions and leave them unused anyways.

TheIssueFairy:
It's ready, by the way, unless there's an issue I don't know about.

PR #2821, opened 2026-02-23
mon_nether_huntsman by Range

I-am-Erk:
#### Summary
Upload from discord: mon_nether_huntsman by Range

#### Content of the change
Adds this sprite
image


Fris0uman:
just wait for Range to give us their github so that we can make them coauthor of the PR to track credits

PR #2863, opened 2026-04-24
[Ultica] Fix rotation on road/dirt/trail corners

ryanaldrich:
#### Summary

@thaelina's recent fixes swapped positions 1 and 3 of every rotation array, but left the base slot unchanged. For most multitile corner subtiles that was correct, because sprites named *_corner_nw seem to convey "the NW outer corner of this shape".

Roads, dirt roads, and forest trails use the opposite convention for some reason. *_corner_nw is the road itself curving through the NW quadrant, meaning the tile connects on N+W.

#### Content of the change
This commit rotates the `corner` subtile arrays to account for this.

#### Testing
I didn't bother taking screenshots, but I validated this in-game. I also double checked things like sidewalks, railroads (turned on the railroad mod to generate them), and fences but those all used the expected conventions. 

#### Additional information
There might be other types like this that I've missed, but I haven't seen any yet.

### DIFF
--- a/gfx/UltimateCataclysm/pngs_normal_32x32/overmap/dirt_road/dirt_road.json
+++ b/gfx/UltimateCataclysm/pngs_normal_32x32/overmap/dirt_road/dirt_road.json
@@ -8,7 +8,7 @@
       { "id": "center", "fg": "dirt_road_center" },
       {
         "id": "corner",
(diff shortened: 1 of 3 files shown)


ryanaldrich:
Closing for now while I learn how composition works. Turns out that the source of the issue was failure to compose the tileset in the first place. The part that makes no sense to me is that these tiles all look correct (when properly composed) regardless of whether or not this commit is applied. I have no idea how that could work, so I'm going to go read up some more.

PR #2868, opened 2026-05-01
More Aftershock furniture sprites

John-Candlebury:
#### Summary
Sprites for modded furniture


Fris0uman:
The sprites fro Retrodays need a full black background and no transparency, also it would be better is the aftershock folder were in a `mod` folder

PR #2873, opened 2026-05-11
misc

CitizensAlias:
misc, readjusted welding tanks to be 20x20 px


Fris0uman:
funnel separation is missing some black underline

CitizensAlias:
fair enough

Fris0uman:
looks like the black pixels are partially transparent, they should be full black I think

CitizensAlias:
I thought that full black outline looked slightly off, but looking in the style guidelines, you explicitly mention 100% black outline, my bad. I fixed it now.
image

PR #2889, opened 2026-06-16
MSXOtto - Fix Freezer Sprites not orientating correctly. Added Vehicle Part Shapes to allow for players to choose what direction for freezer to be placed and temp-assigned sprite to fridges

armaswiss:
#### Summary

I noticed in vehicles, the sprite for the freezers was permanently facing north. The Food Truck Freezer only had 1 shape, meaning you could not choose what direction it was facing.

Due to this, I added vehicle part shapes and expanded the tile_config.json to assign the sprites for each cardinal direction start. Turning the vehicle also has the sprites rotate to remain facing the correct direction in relation to the shape that was chosen. 

A seperate PR will be pushed to CDDA to attach the additional variants to the vehicle_parts.json

#### Content of the change

See above. Added additional tile ID's to match variants added to vehicle_parts and allow for the Food Truck Freezer and Fridge to have 4 shapes (Up/Down/Left/Right) based on a vehicle pointing north

#### Testing

Originally, Freezers always faced north:
image

During Testing, Four Freezers added all facing a unique cardinal direction.
image
Freezers correctly rotated based off vehicle rotation:
image
image
image

### DIFF
--- a/gfx/MShockXotto+/pngs_tall_32x48/furniture/f_fridge/f_fridge.json
+++ b/gfx/MShockXotto+/pngs_tall_32x48/furniture/f_fridge/f_fridge.json
@@ -5,9 +5,9 @@
   ],
   "rotates": true,
   "fg": [
-    "f_fridge",
-    "f_fridge_E",
     "f_fridge_N",
+    "f_fridge_E",
+    "f_fridge",
     "f_fridge_W"
   ],
   "bg": ""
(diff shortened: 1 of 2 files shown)


Fris0uman:
Don't put comments in the json it will cause parsing issues

armaswiss:
> Don't put comments in the json it will cause parsing issues

I just saw this as I was laying down for bed. I believe I have removed all the comments without breaking anything but will have to review once I wake up and have a full desktop!

Thank you for the feedback!

Edit: Now that I'm awake and on a desktop, review and made sure all comments are removed per feedback. Adjusted the code for broken variants to ensure they match with the unbroken tile display.

PR #2905, opened 2026-07-12
Turn the chair and armchair

Number-456:
#### Summary
None

#### Content of the change
An attempt to solve the problem [#86895](https://github.com/CleverRaven/Cataclysm-DDA/issues/86895)

PR #2909, opened 2026-07-27
Fix Dresser E/W Facing

Caelinus:
Fixed the reversed east/west facing for dressers.

#### Summary
Similar fix to #2905.

East/West Facing for Dressers are reversed, this would correct that.

#### Content of the change

Rotation order is N/E/S/W, so having "faceE" in the East position makes it so the sprite faces the wall it is rotating towards. By switching them (as faceN and faceS already are) this will make it face the correct direction.

#### Testing

I repacked the sprites after making the change and it looked correct.

#### Additional information

I did do this a week or more ago, but I cannot see anything in the repo that shows it being fixed already. After seeing #2905, I realized I should probably push this change as well. If it has already been fixed by some other means and I am just super blind, sorry.


Caelinus:
How it currently looks:

oldfaceing


How it looks after the change:

image

PR #2913, opened 2026-08-11
[MSX] Portable Power Station

gettingusedto:
#### Summary
Portable Power Station

#### Content of the change

Basegame
- Portable Power Station (item and appliance)

#### Testing

Screenshot_20260811_103724


Fris0uman:
if ap and vp_ap are exactly the same you should just setup a json file rather than duplicating the sprite

gettingusedto:
oh yeah i forgot to delete the former there because i kinda fallen asleep right after. the ap didnt work anyway

PR #2928, opened 2026-09-13
Tileset stub for MSX portraits

RenechCDDA:
#### Summary
This creates a new "MShockXotto+_Portraits" tileset, for portraits.

#### Content of the change
No actual images, just the files needed and some changes to workflows to accomodate it.

#### Testing
https://github.com/RenechCDDA/CDDA-Tilesets/releases/download/2026-09-13/MshockXotto+_Portraits.zip

#### Additional information

Squash this, I used the web editor 🙃


gfx/MShockXotto+_Portraits/tile_info.json
```diff
@@ -0,0 +1,16 @@
...
+    "height": 128
+  },
+  {
+    "tiles.png": {  }
+  },
+  {
+    "portraits.png": { "sprite_width": 128, "sprite_height": 128, "sprite_offset_x": 0, "sprite_offset_y": 0 }
+  },
+  {
+    "fallback.png": { "fallback": true, "sprite_width": 32, "sprite_height": 32, "sprite_offset_x": 0, "sprite_offset_y": 0 }
```
RenechCDDA, line 14: There doesn't seem to be a way to skip the composer adding or requesting a fallback.png, so this must be included.

We could omit the declaration of fallback.png entirely, BUT since the composer will add fallback.png regardless (or fail if it's not there), we might as well declare it. Additionally, since fallback.png is from a 32x32 sprite sheet, we DO need to say that the sprite_width and sprite_height are 32 here, or else it will search for sprites in the fallback assuming a 512x512 size... which leads to out-of-bounds errors, even though nothing is ever drawn from the fallback! Argh!

TL;DR: Composer demands fallback.png whether we like it or not, easier to just wastefully include it.

RenechCDDA:
I believe this is mergeable even with the CI failure. The CI failure is just saying there's no existing `gfx/MShockXotto+_Portraits` directory on master.

Well, yeah. 'Cause we're adding it.

